PR #5, opened 2022-05-26
fix: bump resource requirements for jenkins

kirederik:


abangser:
This PR was created in order to stabilise the Jenkins deployment which seemed to "flap" on both initialisation and startup.

Turns out that increasing the Swap provided to Docker Desktop to `3` was a much simpler and cleaner solution.

PR #17, opened 2023-07-11
unify system test

kirederik:
each resource request takes 40 ~ 90s on ci. we can speed things up
considerably by creating a single resource covering all test cases.


abangser (approved):
LGTM, I don't think moving it from `it` blocks to `by` blocks changes readability and is actually in many ways more appropriate since it is a single resource action that triggers all of these outcomes 👍

PR #20, opened 2023-07-24
feat: pipeline containers to share the same vols

kirederik:
instead of having the previous container's output becoming the next container's output, this commit makes all containers to have access to the same shared volumes;

this commit also creates the mounts under the `/kratix` directory, since the names `input`, `output`, and `metadata` are not that unique and may clash with whatever paths people may use in their dockerfiles

### DIFF
--- a/lib/pipeline/configure.go
+++ b/lib/pipeline/configure.go
@@ -82,17 +82,7 @@ func configurePipelinePod(rr *unstructured.Unstructured, pipelines []platformv1a
 }
 
 func configurePipelineInitContainers(rr *unstructured.Unstructured, pipelines []platformv1alpha1.Pipeline, rrID string) ([]v1.Container, []v1.Volume) {
-	volumes := []v1.Volume{
-		{Name: "shared-input", VolumeSource: v1.VolumeSource{EmptyDir: &v1.EmptyDirVolumeSource{}}},
-		{Name: "shared-output", VolumeSource: v1.VolumeSource{EmptyDir: &v1.EmptyDirVolumeSource{}}},
-		{Name: "shared-metadata", VolumeSource: v1.VolumeSource{EmptyDir: &v1.EmptyDirVolumeSource{}}},
-	}
-	volumeMounts := []v1.VolumeMount{
-		{MountPath: "/kratix/input", Name: "shared-input", ReadOnly: true},
-		{MountPath: "/kratix/output", Name: "shared-output"},
-		{MountPath: "/kratix/metadata", Name: "shared-metadata"},
-	}
-
+	volumes, volumeMounts := pipelineVolumes()
 	readerContainer := readerContainer(rr, "shared-input")
 	containers := []v1.Container{
 		readerContainer,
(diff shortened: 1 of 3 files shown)


lib/pipeline/delete.go
```diff
@@ -36,25 +34,29 @@ func NewDeletePipeline(rr *unstructured.Unstructured, pipelines []platformv1alph
...
-	readerContainer, readerVolume := readerContainerAndVolume(rr)
-	containers := []v1.Container{readerContainer}
-	volumes := []v1.Volume{readerVolume}
+	volumes := []v1.Volume{
+		{Name: "shared-input", VolumeSource: v1.VolumeSource{EmptyDir: &v1.EmptyDirVolumeSource{}}},
+		{Name: "shared-output", VolumeSource: v1.VolumeSource{EmptyDir: &v1.EmptyDirVolumeSource{}}},
+		{Name: "shared-metadata", VolumeSource: v1.VolumeSource{EmptyDir: &v1.EmptyDirVolumeSource{}}},
+	}
+	volumeMounts := []v1.VolumeMount{
+		{MountPath: "/kratix/input", Name: "shared-input", ReadOnly: true},
```
aclevername, line 43: read only 🤔 seems reasonable

lib/pipeline/configure.go
```diff
@@ -81,32 +80,29 @@ func configurePipelinePod(rr *unstructured.Unstructured, pipelines []platformv1a
...
+	volumes := []v1.Volume{
+		{Name: "shared-input", VolumeSource: v1.VolumeSource{EmptyDir: &v1.EmptyDirVolumeSource{}}},
+		{Name: "shared-output", VolumeSource: v1.VolumeSource{EmptyDir: &v1.EmptyDirVolumeSource{}}},
+		{Name: "shared-metadata", VolumeSource: v1.VolumeSource{EmptyDir: &v1.EmptyDirVolumeSource{}}},
+	}
+	volumeMounts := []v1.VolumeMount{
+		{MountPath: "/kratix/input", Name: "shared-input", ReadOnly: true},
+		{MountPath: "/kratix/output", Name: "shared-output"},
+		{MountPath: "/kratix/metadata", Name: "shared-metadata"},
+	}
```
aclevername, line 92: duplicated in configure and delete, shared piece of code for both maybe?

lib/pipeline/shared.go
```diff
@@ -111,15 +111,11 @@ func readerContainerAndVolume(rr *unstructured.Unstructured) (v1.Container, v1.V
 		Image:   "bitnami/kubectl:1.20.10",
 		Command: []string{"sh", "-c", resourceRequestCommand},
 		VolumeMounts: []v1.VolumeMount{
-			{
-				MountPath: "/output",
-				Name:      "vol0",
-			},
+			{MountPath: "/output", Name: volumeName},
```
aclevername, line 114: should we update ours to use the same pattern just for consistency? Tiny thing but I could imagine future us getting mixed up

aclevername (approved):
lgtm, only minor comments

PR #21, opened 2023-08-18
feat: run a new pipeline job on updates

kirederik:
the workplacement controller is not yet responding to the updated work, so the pipeline outcome won't actually make to the destination just yet

### DIFF
--- a/controllers/dynamic_resource_request_controller.go
+++ b/controllers/dynamic_resource_request_controller.go
@@ -18,7 +18,6 @@ package controllers
 
 import (
 	"context"
-	"time"
 
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -30,25 +29,21 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/syntasso/kratix/api/v1alpha1"
 	"github.com/syntasso/kratix/lib/pipeline"
+	"github.com/syntasso/kratix/lib/resourceutil"
 	batchv1 "k8s.io/api/batch/v1"
-	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
-	conditionsutil "sigs.k8s.io/cluster-api/util/conditions"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/yaml"
 )
 
 const (
-	workFinalizer              = kratixPrefix + "work-cleanup"
-	workflowsFinalizer         = kratixPrefix + "workflows-cleanup"
-	deleteWorkflowsFinalizer   = kratixPrefix + "delete-workflows"
-	PipelineCompletedCondition = clusterv1.ConditionType("PipelineCompleted")
+	workFinalizer            = kratixPrefix + "work-cleanup"
+	workflowsFinalizer       = kratixPrefix + "workflows-cleanup"
+	deleteWorkflowsFinalizer = kratixPrefix + "delete-workflows"
 )
 
 var rrFinalizers = []string{workFinalizer, workflowsFinalizer, deleteWorkflowsFinalizer}
@@ -79,7 +74,7 @@ func (r *dynamicResourceRequestController) Reconcile(ctx context.Context, req ct
 		return ctrl.Result{}, nil
 	}
 
-	logger := r.log.WithValues("uid", r.uid, r.promiseIdentifier, req.NamespacedName)
+	logger := r.log.WithValues("uid", r.uid, "promiseID", r.promiseIdentifier, "namespace", req.NamespacedName)
 	resourceRequestIdentifier := fmt.Sprintf("%s-%s-%s", r.promiseIdentifier, req.Namespace, req.Name)
 
 	rr := &unstructured.Unstructured{}
@@ -103,29 +98,35 @@ func (r *dynamicResourceRequestController) Reconcile(ctx context.Context, req ct
 		return addFinalizers(ctx, r.Client, rr, []string{workFinalizer, workflowsFinalizer, deleteWorkflowsFinalizer}, logger)
 	}
 
-	if !r.hasCondition(PipelineCompletedCondition, rr) {
-		r.setStatus(rr, logger, "message", "Pending")
-		r.setCondition(clusterv1.Condition{
-			Type:               PipelineCompletedCondition,
-			Status:             v1.ConditionFalse,
-			Message:            "Pipeline has not completed",
-			Reason:             "PipelineNotCompleted",
-			LastTransitionTime: metav1.NewTime(time.Now()),
-		}, rr, logger)
+	pipelineJobs, err := r.getPipelineJobs(resourceRequestIdentifier, rr.GetNamespace(), logger)
+	if err != nil {
+		logger.Info("Failed getting pipeline jobs", "error", err)
+		return slowRequeue, nil
+	}
+
+	if len(pipelineJobs) != 0 {
+		if resourceutil.IsThereAPipelineRunning(logger, pipelineJobs) {
+			return slowRequeue, nil
+		}
+
+		if !resourceutil.IsAnUpdate(logger, rr, pipelineJobs) {
+			return ctrl.Result{}, nil
+		}
+	}
+
+	if resourceutil.LastPipelineCompleted(rr) {
+		r.setStatus(rr, logger, "message", "Update Pending")
+		resourceutil.MarkPipelineAsRunning(logger, rr)
 		return ctrl.Result{}, r.Client.Status().Update(ctx, rr)
 	}
 
-	//check if the pipeline has already been created. If it has exit out. All the lines
-	//below this call are for the one-time creation of the pipeline.
-	created, err := r.configurePipelineHasBeenCreated(resourceRequestIdentifier, rr.GetNamespace(), logger)
-	if err != nil {
-		return ctrl.Result{}, err
+	if !resourceutil.HasPipelineCompletedCondition(rr) {
+		r.setStatus(rr, logger, "message", "Pending")
+		resourceutil.MarkPipelineAsRunning(logger, rr)
+		return ctrl.Result{}, r.Client.Status().Update(ctx, rr)
 	}
 
-	if created {
-		logger.Info("Cannot execute update on pre-existing pipeline for Promise resource request " + resourceRequestIdentifier)
-		return ctrl.Result{}, nil
-	}
+	logger.Info("Triggering pipeline", "resourceRequest", resourceRequestIdentifier)
 
 	resources, err := pipeline.NewConfigurePipeline(
 		rr,
@@ -224,16 +225,16 @@ func (r *dynamicResourceRequestController) getDeletePipeline(ctx context.Context
 	return &jobs[0], nil
 }
 
-func (r *dynamicResourceRequestController) configurePipelineHasBeenCreated(resourceRequestIdentifier, namespace string, logger logr.Logger) (bool, error) {
+func (r *dynamicResourceRequestController) getPipelineJobs(rrID, namespace string, logger logr.Logger) ([]batchv1.Job, error) {
 	jobs, err := r.getJobsWithLabels(
-		pipeline.ConfigurePipelineLabels(resourceRequestIdentifier, r.promiseIdentifier),
+		pipeline.ConfigurePipelineLabels(rrID, r.promiseIdentifier),
 		namespace,
 		logger,
 	)
 	if err != nil {
-		return false, err
+		return nil, err
 	}
-	return len(jobs) > 0, nil
+	return jobs, nil
 }
 
 func (r *dynamicResourceRequestController) getJobsWithLabels(jobLabels map[string]string, namespace string, logger logr.Logger) ([]batchv1.Job, error) {
@@ -319,18 +320,6 @@ func (r *dynamicResourceRequestController) deleteWorkflows(ctx context.Context, 
 	}
 
 	return nil
-}
-
-func (r *dynamicResourceRequestController) hasCondition(conditionType clusterv1.ConditionType, rr *unstructured.Unstructured) bool {
-	getter := conditionsutil.UnstructuredGetter(rr)
-	condition := conditionsutil.Get(getter, conditionType)
-	return condition != nil
-}
-
-func (r *dynamicResourceRequestController) setCondition(condition clusterv1.Condition, rr *unstructured.Unstructured, logger logr.Logger) {
-	setter := conditionsutil.UnstructuredSetter(rr)
-	conditionsutil.Set(setter, &condition)
-	logger.Info("set conditions", "condition", condition.Type, "value", condition.Status)
 }
 
 func (r *dynamicResourceRequestController) setStatus(rr *unstructured.Unstructured, logger logr.Logger, statuses ...string) {
--- a/controllers/dynamic_resource_request_controller.go
+++ b/controllers/dynamic_resource_request_controller.go
@@ -31,6 +31,7 @@ import (
 	"github.com/syntasso/kratix/lib/pipeline"
 	"github.com/syntasso/kratix/lib/resourceutil"
 	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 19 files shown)


lib/hash/hash_test.go
```diff
@@ -0,0 +1,66 @@
...
+	Describe("ComputeHash", func() {
+		const expectedHash = "9bb58f26192e4ba00f01e2e7b136bbd8"
+
+		It("hashes the inpuy Object Spec", func() {
+			actualHash, err := hash.ComputeHash(input)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(actualHash).To(Equal(expectedHash))
+		})
+
+		It("remains constant if the request spec hasn't changed", func() {
```
aclevername, line 43: worth adding labels to the things we trigger changes on? maybe overkill now but could see it in the future

kirederik: you mean adding the spec diff? 🤔

aclevername: I mean if someone adds a new metadata.label as the only change to a RR it may be worth retriggering the pipeline

kirederik: gotcha!

i thought about it, and you are right, i can see as something we will potentially need. however, at the same time, keeping it spec only simplifies the code a bit -- and we can always add triggering on other things in the future?

aclevername: sounds reasonable 💯

controllers/dynamic_resource_request_controller.go
```diff
@@ -103,30 +98,36 @@ func (r *dynamicResourceRequestController) Reconcile(ctx context.Context, req ct
...
-	//below this call are for the one-time creation of the pipeline.
-	created, err := r.configurePipelineHasBeenCreated(resourceRequestIdentifier, rr.GetNamespace(), logger)
-	if err != nil {
-		return ctrl.Result{}, err
+	if len(pipelineJobs) != 0 {
+		if resourceutil.IsThereAPipelineRunning(logger, pipelineJobs) {
+			return slowRequeue, nil
+		}
+
+		if !resourceutil.IsAnUpdate(logger, rr, pipelineJobs) {
```
aclevername, line 112: nitpick: we only call this in the context of checking the negative rather than the positive, maybe inverse the func name?

I'm also struggling to understand what the function does from the name 🤔  is it more `isLastDeployedJobUsingCurrentSpec` (but better)

controllers/util.go
```diff
@@ -78,7 +78,7 @@ func addFinalizers(ctx context.Context, client client.Client, resource client.Ob
 	if err := client.Update(ctx, resource); err != nil {
 		return defaultRequeue, err
 	}
-	return fastRequeue, nil
+	return ctrl.Result{}, nil
```
aclevername, line 81: 🤔 is this because issuing the update the line before triggers the controller to reconcile anyway?

lib/hash/hasher.go
```diff
@@ -0,0 +1,17 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func ComputeHash(input *unstructured.Unstructured) (string, error) {
+	spec, err := json.Marshal(input.Object["spec"])
+	if err != nil {
+		return "", err
+	}
+	return fmt.Sprintf("%x", md5.Sum(spec)), nil
+}
```
aclevername, line 17: The approach LGTM, one interesting this is that something similar happens in k8s controllers under the hood to generate a [pod-template-hash](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#pod-template-hash-label), here's how they do it 
https://github.com/kubernetes/kubernetes/blob/ba10ee7671099254db18f9fe9e779ef6339e935b/pkg/controller/controller_utils.go#L1198-L1213

not sure if we need to mirror the complexity they added, annoyingly we can't use the function since it only accepts a pod spec not a generic map 🙃

lib/pipeline/labels.go
```diff
@@ -39,3 +41,11 @@ func (p pipelineLabels) WithPipelineType(pipelineType string) pipelineLabels {
...
 	return p
 }
+
+func (p pipelineLabels) WithRequestSHA(requestSHA []string) pipelineLabels {
+	if len(requestSHA) == 0 {
+		return p
+	}
+	p[KratixResourceHashLabel] = requestSHA[0]
+	return p
+}
```
aclevername, line 51: why is it a list of sha 🤔

lib/resourceutil/conditions.go
```diff
@@ -0,0 +1,22 @@
+package resourceutil
```
aclevername, line 1: nice

test/system/system_test.go
```diff
@@ -137,6 +137,40 @@ var _ = Describe("Kratix", func() {
 				})
 			})
 
+			PWhen("an existing resource request is updated", func() {
```
aclevername, line 140: P

test/system/system_test.go
```diff
@@ -137,6 +137,40 @@ var _ = Describe("Kratix", func() {
...
+					By("redeploying the contents of /kratix/output to the worker destination", func() {
+						Eventually(func() string {
+							return worker.kubectl("get", "namespace")
+						}, "10s").Should(
+							SatisfyAll(
+								Not(ContainSubstring(oldNamespaceName)),
+								ContainSubstring(newNamespaceName),
+							),
+						)
+					})
```
aclevername, line 171: worth adding that the old namespace is deleted as well?

lib/resourceutil/util.go
```diff
@@ -0,0 +1,97 @@
...
+
+	lastRequestHash := lastPipelineJob.GetLabels()[pipeline.KratixResourceHashLabel]
+	currentRequestHash, err := hash.ComputeHash(rr)
+	if err != nil {
+		logger.Info("Cannot determine if the request is an update. Requeueing", "reason", err.Error())
+		return false
+	}
+
+	return lastRequestHash != currentRequestHash
+}
```
aclevername, line 72: should this live here? It feels like really heavy business logic when compared to the other functions that are in this package. Not a hill I'm going to die on, but this func stands out compared to the rest

lib/resourceutil/util.go
```diff
@@ -0,0 +1,97 @@
...
+	if len(jobs) == 0 {
+		return false
+	}
+
+	var incompleteJobs bool
+	for _, job := range jobs {
+		for _, condition := range job.Status.Conditions {
+			switch condition.Type {
+			// TODO: deal with other conditions; what should kratix do if
+			// the job is in a failed or suspended state?
```
aclevername, line 84: yeah this feels like we will get into a pretty frozen state with the code as is today, maybe if a Job is stalled/suspended we ignore it

controllers/dynamic_resource_request_controller.go
```diff
@@ -103,30 +98,36 @@ func (r *dynamicResourceRequestController) Reconcile(ctx context.Context, req ct
...
+		return ctrl.Result{}, r.Client.Status().Update(ctx, rr)
 	}
 
+	if !resourceutil.HasPipelineCompletedCondition(rr) {
+		r.setStatus(rr, logger, "message", "Pending")
+		resourceutil.MarkPipelineAsRunning(logger, rr)
+		return ctrl.Result{}, r.Client.Status().Update(ctx, rr)
+	}
+
+	logger.Info("Triggering pipeline", "resourceRequest", resourceRequestIdentifier)
```
aclevername, line 129: these two checks are making my head hurt 😅 

is the 1st more `if PipelineIsNotMarkedasUpdatePending; then; markAsUpdatePending; done`

unsure what the 2nd is, particularly in the context of the first 🤔

controllers/dynamic_resource_request_controller.go
abangser: Agreed it would help clarify with a slightly different name. Another option could be `RequiresAnUpdate`.

controllers/util.go
kirederik: yes

lib/hash/hasher.go
kirederik: yeah, i look at it (that's why the code I wrote is very similarly named) and i just can't grasp why so complicated; my feeling is that pod hash collision may be more likely (lots of pods, lots of pods that have the same spec). Went for the simpler approach (md5) until we learn why that's bad

lib/pipeline/labels.go
kirederik: this function is called by `ConfigurePipelineLabels` , which sometimes is called without a request sha. i could change it to not be a slice, but would have to add the `if len...` on that function (which is maybe better?)

lib/hash/hasher.go
aclevername: > can't grasp why so complicated

same. Sounds good

lib/pipeline/labels.go
aclevername: ahh I see.

> but would have to add the if len... on that function (which is maybe better?)

 I think its worth changing, its bit more verbose for a clearer func def

lib/resourceutil/util.go
kirederik: i don't really know if i need to sort the jobs, just playing it safe here; i can see moving this out of here, but also don't care too much 👀 apart from the sort, its really just comparing the hashes.

lib/resourceutil/util.go
```diff
@@ -0,0 +1,97 @@
...
+	currentRequestHash, err := hash.ComputeHash(rr)
+	if err != nil {
+		logger.Info("Cannot determine if the request is an update. Requeueing", "reason", err.Error())
+		return false
+	}
+
+	return lastRequestHash != currentRequestHash
+}
+
+func IsThereAPipelineRunning(logger logr.Logger, jobs []batchv1.Job) bool {
```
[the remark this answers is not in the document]
kirederik: i think i disagree with that one. Yes, you are right, that's what we are checking in the function. However, from the controller code, what we are trying to determine is whether there's a pipeline currently running on that resource request. Reading:


```
resourceutil.AreAnyJobsRunning(...)
```

on the controller makes it, imo, less clear.

test/system/system_test.go
kirederik: it's not passing, since the updates are not completed 🤠

test/system/system_test.go
kirederik: thats line 167?

controllers/dynamic_resource_request_controller.go
kirederik: my reasoning behind the name of this functions were based on the package `resourceutil`. From a resource request point of view, what we are trying to determine whether:

- is there a pipeline running for the resource request?
- is this resource request an update?

You are both right that the suggested names better explain what the function is _doing_, but i'd say they are worse in explaining _why_.

controllers/dynamic_resource_request_controller.go
```diff
@@ -103,30 +98,36 @@ func (r *dynamicResourceRequestController) Reconcile(ctx context.Context, req ct
...
+		r.setStatus(rr, logger, "message", "Update Pending")
+		resourceutil.MarkPipelineAsRunning(logger, rr)
+		return ctrl.Result{}, r.Client.Status().Update(ctx, rr)
 	}
 
+	if !resourceutil.HasPipelineCompletedCondition(rr) {
+		r.setStatus(rr, logger, "message", "Pending")
+		resourceutil.MarkPipelineAsRunning(logger, rr)
+		return ctrl.Result{}, r.Client.Status().Update(ctx, rr)
+	}
```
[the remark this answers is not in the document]
kirederik: good question!

If it gets to the line 117, we have determine a pipeline needs to be created.

Those two checks are basically trying to determine what should we update the condition/status message with. 

Line 117 checks if the resource already has a `PipelineCompleted: truer` condition. If so, update the message as `update` and mark the job as running.

line 123 checks if there's even a `PipelineCompleted` condition. If not, add one.

Maybe the issue here is poorly named functions... 🤔

controllers/dynamic_resource_request_controller.go
kirederik: see above

lib/resourceutil/util.go
```diff
@@ -0,0 +1,97 @@
...
+		Type:               PipelineCompletedCondition,
+		Status:             v1.ConditionTrue,
+		Message:            "Pipeline completed",
+		Reason:             "PipelineExecutedSuccessfully",
+		LastTransitionTime: metav1.NewTime(time.Now()),
+	})
+	logger.Info("set conditions", "condition", PipelineCompletedCondition, "value", v1.ConditionTrue)
+}
+
+func IsAnUpdate(logger logr.Logger, rr *unstructured.Unstructured, jobs []batchv1.Job) bool {
```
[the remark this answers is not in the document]
kirederik: see https://github.com/syntasso/kratix/pull/21/files/e6aecf8e873a4aac9f1fc6e15bffccda471e6b53#r1300003109

lib/resourceutil/util.go
kirederik: yeah right now is not gonna trigger an update until the pipeline completes successfully. i don't think we need to fix this now to merge this in though (we do need to think about it and fix it prior to "releasing")

lib/resourceutil/util_test.go
```diff
@@ -0,0 +1,220 @@
...
+							"kratix-resource-hash": originalHash,
+						},
+					},
+					Status: completedStatus,
+				},
+			}
+			Expect(resourceutil.IsThereAPipelineRunning(logger, jobs)).To(BeFalse())
+		})
+
+		It("returns false if there's a job with the JobCompleted: False condition", func() {
```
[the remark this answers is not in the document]
kirederik: nice catch!

controllers/dynamic_resource_request_controller.go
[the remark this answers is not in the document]
kirederik: wdyt of this in lieu of the entire code block?


```golang
	switch resourceutil.GetPipelineCompletedConditionStatus(rr) {
	case corev1.ConditionTrue:
		r.setStatus(rr, logger, "message", "Update Pending")
		resourceutil.MarkPipelineAsRunning(logger, rr)
		return ctrl.Result{}, r.Client.Status().Update(ctx, rr)
	case corev1.ConditionUnknown:
		r.setStatus(rr, logger, "message", "Pending")
		resourceutil.MarkPipelineAsRunning(logger, rr)
		return ctrl.Result{}, r.Client.Status().Update(ctx, rr)
	}
```

controllers/dynamic_resource_request_controller.go
aclevername: > is this resource request an update?

I think I read `resourceutil` to be k8s resources, which I'm only now grokking as resource (request) utils which adds a lot more clarity to me, but `IsAnUpdate` still sounds odd, the negative checking might just be a big nit on my side. 

> You are both right that the suggested names better explain what the function is doing, but i'd say they are worse in explaining why.

🤔 this is a super interesting comment! I think knowing what the function is doing from a function name is a good outcome, it saves the caller from reading the function. But maybe thats more suitable for local functions, and less for a library thats trying to encapsulate the logic (I would still says its not a bad outcome for a library though!). I think using the function name to describe why is super interesting here, but I still think `!resourceutil.IsAnUpdate(logger, rr, pipelineJobs)` hard to read 😅

test/system/system_test.go
aclevername: image

controllers/dynamic_resource_request_controller.go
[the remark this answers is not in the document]
aclevername: > Maybe the issue here is poorly named functions..

I think so. 

It also took me waaay to long to grok that the end output of these two is that you get `Pending` for the first job, and `Update pending` for the rest

lib/resourceutil/util.go
aclevername: 💯

controllers/dynamic_resource_request_controller.go
aclevername: Maybe this function should be in the `pipeline` package? Since its searching through the jobs created by the package it seems reasonable, then @abangser's `RequiresAnUpdate` feels cleaner as `pipeline.RequiresAnUpdate`?

controllers/dynamic_resource_request_controller.go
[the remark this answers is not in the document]
aclevername: > wdyt of this in lieu of the entire code block?

LGTM. I could also live with not having the distinction between `Pending`/`Update Pending` if thats an option, but either are good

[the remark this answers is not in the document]
kirederik: fair, the `update pending` may not add a lot of value tbf... i'll update the code

controllers/dynamic_resource_request_controller.go
kirederik: >I think using the function name to describe why is super interesting here

I think it's preferable!
If tomorrow we decide to determine whether is an update based on a label, or include metadata, or something, we don't need to update the function name... lemme go back to the board and try something considering your feedback 🤔

lib/resourceutil/util.go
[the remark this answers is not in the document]
kirederik: resolving this as we're chatting on the other thread!

aclevername (approved):
:shipit:

PR #28, opened 2023-08-25
Remove old workload files when updating a workplacement

aclevername:


controllers/work_controller_test.go
```diff
@@ -104,6 +104,10 @@ var _ = Context("WorkReconciler.Reconcile()", func() {
 			workPlacementGenID = workPlacementList.Items[0].GetGeneration()
 		})
 
+		AfterEach(func() {
+			deleteWork(work)
+		})
+
```
kirederik, line 110: this `AfterEach` block will run after the `It` in the `When` block below. Moving it to here sounds... odd? i'd be fine moving it up (just after the `BeforeEach`) or as the last thing (as it was before) 🤔

aclevername: I think @abangser moved this before I started working on branch (or I have no memory of this change). Happy to move it back

lib/writers/git.go
```diff
@@ -68,125 +68,139 @@ func NewGitWriter(logger logr.Logger, stateStoreSpec platformv1alpha1.GitStateSt
...
-		"fileName", fileName,
-		"branch", g.gitServer.Branch,
-	)
-	if len(toWrite) == 0 {
-		log.Info("Empty byte[]. Nothing to write to Git")
-		return nil
-	}
-
-	localTmpDir, err := createLocalDirectory()
+func (g *GitWriter) setupLocalDirectoryWithRepo(logger logr.Logger) (string, *git.Repository, *git.Worktree, error) {
```
kirederik, line 71: thats a lot of things being returned 🤠

lib/writers/statestore_writer.go
```diff
@@ -1,6 +1,13 @@
 package writers
 
+import platformv1alpha1 "github.com/syntasso/kratix/api/v1alpha1"
+
+const (
+	DeleteExistingContentsInDir   = true
```
kirederik, line 6: this could be type aliased. but not a big deal

lib/writers/git.go
kirederik: you can also wrap the returns in a struct?

```golang
type gitThing struct {
  localPath string
  repo *git.Repository
  worktree *git.Worktree
}
```

kirederik: not a big deal btw, feel free to ignore

lib/writers/git.go
```diff
@@ -68,125 +68,139 @@ func NewGitWriter(logger logr.Logger, stateStoreSpec platformv1alpha1.GitStateSt
...
+		//worker-cluster/resources/<rr-namespace>/<promise-name>/<rr-name>/foo/bar/baz.yaml
+		worktreeFilePath := filepath.Join(dirInGitRepo, item.Filepath)
+		logger := logger.WithValues(
+			"filepath", worktreeFilePath,
+		)
+
+		///tmp/git-dir/worker-cluster/resources/<rr-namespace>/<promise-name>/<rr-name>/foo/bar/baz.yaml
+		absoluteFilePath := filepath.Join(localTmpDir, worktreeFilePath)
+
+		//We need to protect against paths containg `..`
```
kirederik, line 132: are we protecting against ourselves here? 🤔

lib/writers/git.go
```diff
@@ -68,125 +68,139 @@ func NewGitWriter(logger logr.Logger, stateStoreSpec platformv1alpha1.GitStateSt
...
+		//filepath.Join expands any '../' in the path to the actual, e.g. /tmp/foo/../ resolves to /tmp/
+		//To ensure they can't write to files on disk outside of the tmp git repostiroy we check the absolute path
+		//returned by `filepath.Join` is still contained with the git repository:
+		// Note: This means `../` can still be used, but only if the end result is still contained within the git repository
+		if !strings.HasPrefix(absoluteFilePath, localTmpDir) {
+			logger.Error(nil, "path of file to write is not located within the git repostiory", "absolutePath", absoluteFilePath, "tmpDir", localTmpDir, "repoPath", worktreeFilePath, "path", g.path)
```
kirederik, line 138: `"repoPath", worktreeFilePath` already included in the logger as `filepath`

lib/writers/git.go
```diff
@@ -68,125 +68,139 @@ func NewGitWriter(logger logr.Logger, stateStoreSpec platformv1alpha1.GitStateSt
...
+	defer os.RemoveAll(filepath.Dir(localTmpDir))
 
-	objectFileName := filepath.Join(g.path, fileName)
-	if _, err := worktree.Filesystem.Lstat(objectFileName); err == nil {
-		if _, err := worktree.Remove(objectFileName); err != nil {
-			log.Error(err, "could not remove file from worktree")
+	worktreeFilepath := filepath.Join(g.path, filePath)
+	if _, err := worktree.Filesystem.Lstat(worktreeFilepath); err == nil {
+		if _, err := worktree.Remove(worktreeFilepath); err != nil {
+			logger.Error(err, "could not remove file from worktree")
```
kirederik, line 174: `worktree.Remove` implementation calls `Lstat` prior to deleting, and return the error if `Lstat` errors. I'm assuming that the error is not useful for a check?

lib/writers/s3.go
```diff
@@ -52,59 +52,119 @@ func NewS3Writer(logger logr.Logger, stateStoreSpec platformv1alpha1.BucketState
...
-			logger.Info("Object does not exist yet")
+		ctx := context.Background()
+
+		// Check to see if we already own this bucket (which happens if you run this twice)
+		exists, errBucketExists := b.RepoClient.BucketExists(ctx, b.BucketName)
+		if errBucketExists != nil {
+			logger.Error(errBucketExists, "Could not verify bucket existence with provider")
+		} else if !exists {
+			logger.Info("Bucket provided does not exist (or the provided keys don't have permissions)")
+		}
```
kirederik, line 91: refactor: do we need to test this for each file? or just once before the loop?

lib/writers/s3.go
```diff
@@ -52,59 +52,119 @@ func NewS3Writer(logger logr.Logger, stateStoreSpec platformv1alpha1.BucketState
...
+		reader := bytes.NewReader([]byte(item.Content))
+
+		objStat, err := b.RepoClient.StatObject(ctx, b.BucketName, objectFullPath, minio.GetObjectOptions{})
+		if err != nil {
+			if minio.ToErrorResponse(err).Code == "NoSuchKey" {
+				logger.Info("Object does not exist yet")
+			} else {
+				logger.Error(err, "Error fetching object")
+				return err
+			}
```
kirederik, line 102: nit: i reckon this was already there, but... 😆

```
if minio.ToErrorResponse(err).Code != "NoSuchKey" {
    logger.Error(err, "Error fetching object")
    return err
}
logger.Info("Object does not exist yet")
```

lib/writers/s3.go
```diff
@@ -52,59 +52,119 @@ func NewS3Writer(logger logr.Logger, stateStoreSpec platformv1alpha1.BucketState
...
 		}
+	}()
+
+	errorCh := b.RepoClient.RemoveObjects(ctx, b.BucketName, objectsCh, minio.RemoveObjectsOptions{})
+
+	// Print errors received from RemoveObjects API
+	var errors []error
+	for e := range errorCh {
+		logger.Error(e.Err, "Failed to remove object", "objectName", e.ObjectName)
+		errors = append(errors, e.Err)
```
kirederik, line 141: refactor: it looks like we are using this only to count the errors? maybe an `int`?

lib/writers/git.go
aclevername: I think anyone who creates works/workplacements can pass in `..` so its a reasonable check (this was pre-existing before this PR)

lib/writers/git.go
aclevername: ohh interesting I didn't know that. I think its useful as is to distinguish if the error from `remove` is because it doesn't exist vs something else

lib/writers/s3.go
aclevername: good spot, moved outside

kirederik (approved):

Issue #27, opened 2023-08-24
[Bug] Reinstalling a Promise where the API Plural changes prevents kratix from reconciling resource requests


aclevername:
## Steps to reproduce:
1. Install a promise
2. Delete the promise
3. Update the promise.yaml so the `plural` of the API changes
4. Install the promise
5. Make a resource request

We want a pipeline to get triggered for the request, but in reality nothing happens. Kratix is misbehaving.

## Underlying Issue

Kubernetes clients cache a history of known CRDs, for example:
<img width="1139" alt="Screenshot 2023-08-24 at 12 52 12" src="https://github.com/syntasso/kratix/assets/19669095/b308c98a-0ff2-4705-918c-5f22822d621d">

If you delete the CRD and create it again, with the plural changed the cache doesn't get updated, preventing the `get` from working:
<img width="975" alt="Screenshot 2023-08-24 at 12 53 34" src="https://github.com/syntasso/kratix/assets/19669095/a2cbbcf6-11f1-41be-824d-bb5ff67fe572">

This same issue is happening internally in the Kratix controller. We watch for the new CRD created by the Promise, but because we've cached the old plural we are unable to reconcile as we try to fetch the old plural instead of the new.

## Workaround
Restarting the Kratix pod resolves the issue as this resets the cache.

Issue #34, opened 2023-09-08
[Bug] Remove or unserving the original version in the API prevents kratix from reconciling resource requests


aclevername:
## Steps to reproduce:
1. Install a promise with API  versions `v1` `served: true`
2. Update the promise to also have `v2` `served:true`
3. Update the promise to v1 is either `served: false` or just removed
5. Make a resource request

We want a pipeline to get triggered for the request, but in reality nothing happens. Kratix is misbehaving.

## Underlying Issue

We start the [dynamic controller with a particular GVK](https://github.com/syntasso/kratix/blob/8ab40be7ac83c62006bad72d9e75292b6ff70007/controllers/promise_controller.go#L233-L238), this is the GVK the controller manager watches for. If this Version ever stops being served (either by `served: false` or deletion) the dynamic controller won't be able to watch for the version, and won't ever reconcile any resources.

## Workaround
Restarting the Kratix pod resolves the issue as this resets the version the controller watches.

PR #35, opened 2023-09-18
fix: stop dynamic controller on promise deletion by restarting manager

aclevername:
Fixes 

https://github.com/syntasso/kratix/issues/27
https://github.com/syntasso/kratix/issues/34

### DIFF
--- a/controllers/promise_controller.go
+++ b/controllers/promise_controller.go
@@ -47,6 +47,7 @@ type PromiseReconciler struct {
 	Log                       logr.Logger
 	Manager                   ctrl.Manager
 	StartedDynamicControllers map[string]*dynamicResourceRequestController
+	RestartManager            func()
 }
 
 const (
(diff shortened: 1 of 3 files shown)


controllers/promise_controller.go
```diff
@@ -363,6 +364,7 @@ func (r *PromiseReconciler) deletePromise(ctx context.Context, promise *v1alpha1
 	//temporary fix until https://github.com/kubernetes-sigs/controller-runtime/issues/1884 is resolved
 	//once resolved, delete dynamic controller rather than disable
 	if d, exists := r.StartedDynamicControllers[string(promise.GetUID())]; exists {
+		r.RestartManager()
 		enabled := false
 		d.enabled = &enabled
```
aclevername, line 369: keeping the rest of the "old" implementation so the unit tests don't misbehave

main.go
```diff
@@ -59,80 +60,93 @@ func main() {
...
+			setupLog.Error(err, "unable to set up health check")
+			os.Exit(1)
+		}
+		if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
+			setupLog.Error(err, "unable to set up ready check")
+			os.Exit(1)
+		}
+
+		setupLog.Info("starting manager")
+		err = mgr.Start(ctx)
```
kirederik, line 142: this `err` is never checked; maybe that's what should be checked on the line below?

PR #36, opened 2023-09-18
Promise workflows

aclevername:


controllers/dynamic_resource_request_controller.go
```diff
@@ -311,59 +242,59 @@ func (r *dynamicResourceRequestController) deleteWork(ctx context.Context, resou
...
+		if err := r.Client.Update(o.ctx, resourceRequest); err != nil {
 			return err
 		}
 	}
 
 	return nil
 }
 
-func (r *dynamicResourceRequestController) setStatus(rr *unstructured.Unstructured, logger logr.Logger, statuses ...string) {
+func setStatus(rr *unstructured.Unstructured, logger logr.Logger, statuses ...string) {
```
kirederik, line 297: agree that this doesn't need to be part of the struct; should this (and some of the other functions that got removed from the struct) move to one of the shared libs 🤔? those package-global functions can get hard to follow/find sometimes

controllers/promise_controller.go
```diff
@@ -51,6 +54,7 @@ type PromiseReconciler struct {
 }
 
 const (
+	kratixPlatformSystemNamespace                      = "kratix-platform-system"
```
kirederik, line 57: thos constant is already defined here https://github.com/syntasso/kratix/blob/a43e0b6c12c59e8733ccd58ff4742031c98e5601/api/v1alpha1/groupversion_info.go#L39

controllers/promise_controller.go
```diff
@@ -174,6 +188,54 @@ func (r *PromiseReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
 }
 
+func (r *PromiseReconciler) reconcileDependencies(o opts, promise *v1alpha1.Promise, configurePipeline []v1alpha1.Pipeline) (*ctrl.Result, error) {
+	if len(promise.Spec.Workflows.Promise.Configure) == 0 {
+		o.logger.Info("Promise does not contain workflows.promise.configure, applying dependencies directly")
+		if err := r.applyWorkResourceForDependencies(o, promise); err != nil {
+			o.logger.Error(err, "Error creating Works")
+			return nil, err
+		}
+	} else {
```
kirederik, line 198:
```suggestion
	if len(promise.Spec.Workflows.Promise.Configure) == 0 {
		o.logger.Info("Promise does not contain workflows.promise.configure, applying dependencies directly")
		if err := r.applyWorkResourceForDependencies(o, promise); err != nil {
			o.logger.Error(err, "Error creating Works")
			return nil, err
		}
        return nil, nil
	}
    // else gets dropped
```

controllers/promise_controller.go
```diff
@@ -174,6 +188,54 @@ func (r *PromiseReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		requeue, err := ensurePipelineIsReconciled(jobOpts)
+		if err != nil {
+			return nil, err
+		}
+
+		if requeue != nil {
+			return requeue, nil
+		}
+	}
+	return nil, nil
```
kirederik, line 236: it looks like you can just return `requeue` and `err`?

```suggestion
	return ensurePipelineIsReconciled(jobOpts)
```

controllers/promise_controller.go
```diff
@@ -600,31 +696,53 @@ func (r *PromiseReconciler) applyWorkResourceForDependencies(ctx context.Context
...
 	return nil
 }
 
-func (r *PromiseReconciler) generatePipelines(promise *v1alpha1.Promise, logger logr.Logger) ([]v1alpha1.Pipeline, []v1alpha1.Pipeline, error) {
-	var configurePipelines []v1alpha1.Pipeline
+type promisePipelines struct {
+	DeleteResource    []v1alpha1.Pipeline
+	ConfigureResource []v1alpha1.Pipeline
+	ConfigurePromise  []v1alpha1.Pipeline
+}
```
kirederik, line 707: suggestion: move to `utils.go`?

controllers/promise_controller.go
```diff
@@ -600,31 +696,53 @@ func (r *PromiseReconciler) applyWorkResourceForDependencies(ctx context.Context
...
+	var deleteResourcePipelines []v1alpha1.Pipeline
 	for _, pipeline := range promise.Spec.Workflows.Resource.Delete {
 		p, err := generatePipeline(pipeline, logger)
 		if err != nil {
-			return nil, nil, err
+			return promisePipelines{}, err
 		}
-		deletePipelines = append(deletePipelines, p)
+		deleteResourcePipelines = append(deleteResourcePipelines, p)
 	}
```
kirederik, line 736: could `generatePipeline` receive the unstructured array and return the structured pipeline? we kinda have the same for three times here.

controllers/util.go
```diff
@@ -17,32 +19,122 @@ import (
...
+		return &fastRequeue, createConfigurePipeline(j)
+	}
+
+	existingPipelineJob, err := resourceutil.PipelineExists(j.logger, j.obj, pipelineJobs)
+	if err != nil {
+		return &slowRequeue, nil
+	}
+
+	if resourceutil.IsThereAPipelineRunning(j.logger, pipelineJobs) {
+		for _, job := range resourceutil.SuspendablePipelines(j.logger, pipelineJobs) {
```
kirederik, line 67: 🤔: is this a check we only wanna do if there's a pipeline running? wouldn't we want to suspend all pipelines that are not for the most recent spec?

lib/pipeline/configure.go
```diff
@@ -5,27 +5,32 @@ import (
...
 )
 
 const (
 	kratixConfigureOperation = "configure"
 )
 
-func NewConfigurePipeline(
+func NewConfigureResource(
 	rr *unstructured.Unstructured,
+	//TODO change to plural string
```
kirederik, line 27: 👀

lib/pipeline/configure.go
```diff
@@ -171,3 +225,10 @@ func metadataAndSchedulingVolumes(configMapName string) []v1.Volume {
 		},
 	}
 }
+
+func providedOrDefaultName(providedName string) string {
+	if providedName == "" {
+		return "default-container-name"
```
kirederik, line 231: would this cause multiple containers to have the same name?

lib/pipeline/shared.go
```diff
@@ -115,21 +157,24 @@ func destinationSelectorsConfigMap(resources PipelineArgs, destinationSelectors
...
 
-func readerContainer(rr *unstructured.Unstructured, volumeName string) v1.Container {
-	resourceKindNameNamespace := fmt.Sprintf("%s.%s %s --namespace %s",
-		strings.ToLower(rr.GetKind()), rr.GroupVersionKind().Group, rr.GetName(), rr.GetNamespace())
+func readerContainer(obj *unstructured.Unstructured, volumeName string) v1.Container {
+	namespace := obj.GetNamespace()
+	if namespace == "" {
+		// if namespace is empty it means its a unnamespaced resource, so providing
+		// any value is valid for kubectl
+		namespace = "kratix-platform-system"
```
kirederik, line 165: fix: use the constant

lib/resourceutil/util.go
```diff
@@ -49,9 +49,9 @@ func MarkPipelineAsCompleted(logger logr.Logger, obj *unstructured.Unstructured)
 	logger.Info("set conditions", "condition", PipelineCompletedCondition, "value", v1.ConditionTrue)
 }
 
-func PipelineForRequestExists(logger logr.Logger, rr *unstructured.Unstructured, jobs []batchv1.Job) (bool, error) {
+func PipelineExists(logger logr.Logger, obj *unstructured.Unstructured, jobs []batchv1.Job) (*batchv1.Job, error) {
```
kirederik, line 52: 🤔 i found this renaming a bit confusing
i undestand it's no longer used only for requests, but without the qualifying `ForX`, it no longer conveys the sense that "a pipeline for this spec" exists. unsure how to name it though

work-creator/pipeline/work_creator.go
```diff
@@ -21,9 +22,13 @@ type WorkCreator struct {
 	K8sClient client.Client
 }
 
-func (w *WorkCreator) Execute(rootDirectory, promiseName, namespace, resourceName string) error {
+func (w *WorkCreator) Execute(rootDirectory, promiseName, namespace, resourceName string, addPromiseDependencies bool) error {
 	identifier := fmt.Sprintf("%s-%s", promiseName, resourceName)
 
+	if namespace == "" {
+		namespace = "kratix-platform-system"
```
kirederik, line 29: use the constant?

work-creator/pipeline/cmd/main.go
```diff
@@ -17,11 +17,13 @@ func main() {
 	var promiseName string
 	var namespace string
 	var resourceName string
+	var addPromiseDependencies bool
 
 	flag.StringVar(&inputDirectoy, "input-directory", "", "Absolute path to directory containing yaml documents required to build Work")
 	flag.StringVar(&promiseName, "promise-name", "", "Name of the promise")
-	flag.StringVar(&namespace, "namespace", "default", "Namespace")
+	flag.StringVar(&namespace, "namespace", "kratix-platform-system", "Namespace")
```
kirederik, line 24: use the constant?

config/manager/manager.yaml
```diff
@@ -28,7 +28,7 @@ spec:
       - command:
         - /manager
         args:
-        - --leader-elect
+        - --leader-elect=false
```
kirederik, line 31: curious to what caused this thing to appear here

aclevername: lets chat about this in zoom

controllers/dynamic_resource_request_controller.go
aclevername: thats a good shout. I'll see what can be moved out

controllers/promise_controller.go
aclevername: that is what it does 🤔 I will see if anything can be done to reduce the repetition

controllers/util.go
aclevername: We only *need* to suspend if one is running, if none are inflight (all jobs are complete/suspened/failed) its okay not to.

lib/resourceutil/util.go
aclevername: I gave it a go, lmk what you think

controllers/dynamic_resource_request_controller.go
aclevername: moved a bunch, lmk what you think

PR #43, opened 2023-10-16
fix: overwriting status on pipeline re-run

kirederik:
when the pipeline gets retriggered for a particular resource, the
  controller will add a 'message: Pending' status to the resource.

  however, the old code was accidentaly wiping out any other status from
  the object; as a side effect, pipelines that relied on information
  persisted in the `status` field wouldn't have that information
  available.

  this commit changes the `SetStatus` utils to utilise the existing
  resource `status` as a base (if that exists).


abangser (approved):
Nice!

PR #55, opened 2023-12-12
feat: make auth method configurable for buckets, add support for IAM

aclevername:


controllers/util.go
```diff
@@ -199,6 +199,12 @@ func fetchObjectAndSecret(o opts, stateStoreRef client.ObjectKey, stateStore Sta
 		o.logger.Error(err, "unable to fetch resource", "resourceKind", stateStore.GetObjectKind(), "stateStoreRef", stateStoreRef)
 		return nil, err
 	}
+
+	if stateStore.GetSecretRef() == nil {
+		o.logger.Info("No secret ref found, assuming IAM role should be used", "stateStore", stateStoreRef)
```
kirederik, line 204: should this be assumed or should the code verify if the authMethod is `iam`?

aclevername: oops, we should just log nothing and continue

lib/writers/s3.go
```diff
@@ -24,20 +29,35 @@ type S3Writer struct {
...
 
-	secretAccessKey, ok := creds["secretAccessKey"]
-	if !ok {
-		return nil, fmt.Errorf("secretAccessKey not found in secret %s/%s", destination.Namespace, stateStoreSpec.SecretRef.Name)
+	logger.Info("setting up s3 client", "authMethod", stateStoreSpec.AuthMethod, "endpoint", endpoint, "insecure", stateStoreSpec.Insecure)
+	switch stateStoreSpec.AuthMethod {
+	case AuthMethodIAM:
+		opts.Creds = credentials.NewIAM("")
+
+	case "", AuthMethodAccessKey: //used to be optional so lets handle empty as the default
```
kirederik, line 41: alternatively, we could separate and use `fallthrough`:

```go
case "":
  // log?
  fallthrough
case AuthMethodAccessKey:
  // ...
```

kirederik: also, will it ever be empty? won't the crd change automagically set the default for us?

lib/writers/s3_test.go
```diff
@@ -0,0 +1,118 @@
...
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace: "default",
+					Name:      "test",
+				},
+				Spec: platformv1alpha1.DestinationSpec{},
+			}
+
+		})
+
+		Context("secretAccessKey", func() {
```
kirederik, line 39: nit: is this the context for the `authMethod`? should it be `accessKey` instead of `secretAccessKey`?

lib/writers/s3.go
aclevername: i did not know that was even a thing 😱

lib/writers/s3_test.go
aclevername: oops yep. Originally I called the auth method `secretAccessKey`, decided to rename to `accessKey` but missed this! thanks

PR #57, opened 2023-12-15
feat: release to respect promise-defined version

kirederik:
the promise version should be defined in the promise yaml itself.

the promise can define its version by setting the following label:
    kratix.io/promise-version: <version>

* if a promise release sets a version, and the version matches the version in the promise release, all is good
* if a promise release doesn't set a version, its version will be set to the version of the promise found on the url
* if the promise release version doesn't match the promise version found on the url, the promise won't be installed.

[#186673378](https://www.pivotaltracker.com/story/show/186673378)

### DIFF
--- a/controllers/assets/redis-simple-promise-updated.yaml
+++ b/controllers/assets/redis-simple-promise-updated.yaml
@@ -3,6 +3,7 @@ kind: Promise
 metadata:
   name: redis
   labels:
+    kratix.io/promise-version: v1.2.0
     clashing-label: "new-promise-v2-value"
     new-promise-v2-label: "value"
   annotations:
--- a/controllers/promise_controller_test.go
+++ b/controllers/promise_controller_test.go
@@ -52,9 +52,9 @@ var _ = Describe("PromiseController", func() {
 			})
 		})
 
-		When("the promise is created via a PromiseRelease", func() {
+		When("the promise contains a version label", func() {
 			BeforeEach(func() {
-				promise.Labels["kratix.io/promise-release-version"] = "v1.2.3"
+				promise.Labels["kratix.io/promise-version"] = "v1.2.3"
 				Expect(fakeK8sClient.Create(ctx, promise)).To(Succeed())
 
 				result, err := reconcile(&reconciler, promise, &opts{singleReconcile: true})
@@ -62,7 +62,7 @@ var _ = Describe("PromiseController", func() {
 				Expect(result).To(Equal(ctrl.Result{}))
 			})
 
-			It("persists the value labels in the Status", func() {
+			It("it gets its status updated with the version", func() {
 				promise := fetchPromise(promiseName)
 				Expect(promise.Status.Version).To(Equal("v1.2.3"))
 			})
--- a/controllers/promiserelease_controller.go
+++ b/controllers/promiserelease_controller.go
@@ -96,6 +96,10 @@ func (r *PromiseReleaseReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		if err != nil {
 			return ctrl.Result{}, fmt.Errorf("failed to fetch promise from url: %w", err)
 		}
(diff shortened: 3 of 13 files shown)


controllers/assets/redis-simple-promise-updated.yaml
```diff
@@ -3,6 +3,7 @@ kind: Promise
 metadata:
   name: redis
   labels:
+    kratix.io/promise-version: v1.2.0
```
aclevername, line 6:
```suggestion
    kratix.io/promise/version: v1.2.0
```

thoughts?

controllers/promise_controller_test.go
```diff
@@ -52,17 +52,17 @@ var _ = Describe("PromiseController", func() {
...
+				promise.Labels["kratix.io/promise-version"] = "v1.2.3"
 				Expect(fakeK8sClient.Create(ctx, promise)).To(Succeed())
 
 				result, err := reconcile(&reconciler, promise, &opts{singleReconcile: true})
 				Expect(err).NotTo(HaveOccurred())
 				Expect(result).To(Equal(ctrl.Result{}))
 			})
 
-			It("persists the value labels in the Status", func() {
+			It("it gets its status updated with the version", func() {
```
aclevername, line 65: why is this the behaviour 🤔 shouldn't we just set the status directly when we create the promise in the promiserelease controller

controllers/promiserelease_controller.go
```diff
@@ -225,3 +228,29 @@ func (r *PromiseReleaseReconciler) reconcileOnInstalledPromise(o opts, promiseRe
...
+	if promiseVersion != promiseRelease.Spec.Version {
+		return false, fmt.Errorf(
+			"version label on promise (%s) does not match version on promise release (%s); refusing to install",
+			promiseVersion,
+			promiseRelease.Spec.Version,
+		)
+	}
+
+	return false, nil
+}
```
aclevername, line 256: do we wan to add support for some sort of `latest`, `any`/blank or `*` option? i imagine that would be useful for dev

controllers/assets/redis-simple-promise-updated.yaml
kirederik: dont find it better enough to change the code, but also wouldn't mind if someone wants to update it :D

controllers/promise_controller_test.go
kirederik: afaik you need to do 2 client calls to set status in the promise controller?

test/system/system_test.go
```diff
@@ -483,7 +483,10 @@ var _ = Describe("Kratix", func() {
...
 		When("a PromiseRelease is installed", func() {
 			BeforeEach(func() {
-				platform.kubectl("apply", "-f", promiseReleasePath)
+				tmpDir, err := os.MkdirTemp(os.TempDir(), "systest")
+				Expect(err).NotTo(HaveOccurred())
+				platform.kubectl("apply", "-f", catAndReplace(tmpDir, promiseReleasePath))
+				os.RemoveAll(tmpDir)
 			})
 
 			It("installs the Promises specified", func() {
```
aclevername, line 492: shall we assert the version is set?

controllers/promiserelease_controller.go
kirederik: i think for dev you may just drop the promise-release altogether.

i think adding support for those would be good though, just not sure we need to do it now (there will be an entire epic for floating versions i suppose)

controllers/promise_controller_test.go
aclevername: yeah, but it feels odd that the label is essentially an override mechanism

aclevername: or is that the behaviour to unluck manual version setting?

aclevername (approved):
few comments but nothing big

PR #59, opened 2023-12-18
chore: migrate Scheduler tests from envtest to fake client

richcooper95:
finishes: #186442323

https://www.pivotaltracker.com/story/show/186442323

Scheduler tests are now fast, and have some additional coverage.

Best reviewed commit-by-commit (see commit messages for more context on each one!).

-----

End result: 27 Scheduler tests take 0.032s to run, and including setup/teardown they take 2.7s ⏱️

```
[Syntassos-MBP] [scheduler-envtest-removal] [kratix] 
  > ginkgo run -r controllers
Running Suite: Controller Suite - /Users/richbartoncooper/Projects/kratix/controllers
=====================================================================================
Random Seed: 1703155204

Will run 27 of 74 specs
SSSSSS•••••••••••••••••••••••••••SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS

Ran 27 of 74 Specs in 0.032 seconds
SUCCESS! -- 27 Passed | 0 Failed | 0 Pending | 47 Skipped
PASS | FOCUSED

Ginkgo ran 1 suite in 2.78743775s
Test Suite Passed
Detected Programmatic Focus - setting exit status to 197
```

### DIFF
--- a/controllers/destination_controller.go
+++ b/controllers/destination_controller.go
@@ -92,7 +92,9 @@ func (r *DestinationReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return defaultRequeue, nil
 	}
 
-	if err := r.Scheduler.ReconcileDestination(); err != nil {
+	// TODO: Instead of this function call, we could watch for changes to the Destination
+	// CRD and triggering all Works to reconcile on every change.
+	if err := r.Scheduler.ReconcileAllDependencyWorks(); err != nil {
 		logger.Error(err, "unable to schedule destination resources")
 		return defaultRequeue, nil
 	}
--- a/controllers/scheduler.go
+++ b/controllers/scheduler.go
@@ -38,8 +38,10 @@ type Scheduler struct {
 	Log    logr.Logger
 }
 
-// Only reconciles Works that are from a Promise Dependency
-func (s *Scheduler) ReconcileDestination() error {
+// Reconciles all Works by scheduling each Work's WorkloadGroups to appropriate
+// Destinations.
+// Only reconciles Works that are from a Promise Dependency.
+func (s *Scheduler) ReconcileAllDependencyWorks() error {
 	works := platformv1alpha1.WorkList{}
 	lo := &client.ListOptions{}
 	if err := s.Client.List(context.Background(), &works, lo); err != nil {
@@ -57,36 +59,10 @@ func (s *Scheduler) ReconcileDestination() error {
 	return nil
 }
 
-func (s *Scheduler) UpdateWorkPlacement(workloadGroup platformv1alpha1.WorkloadGroup, work *platformv1alpha1.Work, workPlacement *platformv1alpha1.WorkPlacement) (bool, error) {
-	misscheduled := true
-	destinationSelectors := resolveDestinationSelectorsForWorkloadGroup(workloadGroup, work)
-	for _, dest := range s.getTargetDestinationNames(destinationSelectors, work) {
-		if dest == workPlacement.Spec.TargetDestinationName {
-			misscheduled = false
-			break
-		}
-	}
-
-	if misscheduled {
-		s.labelWorkplacementAsMisscheduled(workPlacement)
-	}
-
-	workPlacement.Spec.Workloads = workloadGroup.Workloads
-	if err := s.Client.Update(context.Background(), workPlacement); err != nil {
-		s.Log.Error(err, "Error updating WorkPlacement", "workplacement", workPlacement.Name)
-		return false, err
-	}
-
-	if err := s.updateStatus(workPlacement, misscheduled); err != nil {
-		return false, err
-	}
-
-	s.Log.Info("Successfully updated WorkPlacement workloads", "workplacement", workPlacement.Name)
-	return misscheduled, nil
-}
-
+// Reconciles all WorkloadGroups in a Work by scheduling them to Destinations via
+// Workplacements.
 func (s *Scheduler) ReconcileWork(work *platformv1alpha1.Work) ([]string, error) {
-	unscheduable := []string{}
+	unschedulable := []string{}
 	misscheduled := []string{}
 	for _, wg := range work.Spec.WorkloadGroups {
 		schedulingStatus, err := s.reconcileWorkloadGroup(wg, work)
@@ -95,7 +71,7 @@ func (s *Scheduler) ReconcileWork(work *platformv1alpha1.Work) ([]string, error)
 		}
 
 		if schedulingStatus == unscheduledStatus {
-			unscheduable = append(unscheduable, wg.ID)
+			unschedulable = append(unschedulable, wg.ID)
 		}
 
 		if schedulingStatus == misscheduledStatus {
@@ -103,11 +79,11 @@ func (s *Scheduler) ReconcileWork(work *platformv1alpha1.Work) ([]string, error)
 		}
 	}
 
-	if err := s.updateWorkStatus(work, unscheduable, misscheduled); err != nil {
+	if err := s.updateWorkStatus(work, unschedulable, misscheduled); err != nil {
 		return nil, err
 	}
 
-	return unscheduable, s.cleanupDanglingWorkplacements(work)
+	return unschedulable, s.cleanupDanglingWorkplacements(work)
 }
 
 func (s *Scheduler) updateWorkStatus(work *platformv1alpha1.Work, unscheduledWorkloadGroupIDs, missscheduledWorkloadGroupIDs []string) error {
@@ -191,8 +167,10 @@ func (s *Scheduler) cleanupDanglingWorkplacements(work *platformv1alpha1.Work) e
 	return nil
 }
 
-// TODO why pointer for work?
+// Reconciles a WorkloadGroup by scheduling it to a Destination via a Workplacement.
 func (s *Scheduler) reconcileWorkloadGroup(workloadGroup platformv1alpha1.WorkloadGroup, work *platformv1alpha1.Work) (schedulingStatus, error) {
+	// TODO why pointer for work?
+
 	existingWorkplacements, err := s.getExistingWorkPlacementsForWorkloadGroup(work.Namespace, work.Name, workloadGroup)
 	if err != nil {
 		return "", err
@@ -200,11 +178,13 @@ func (s *Scheduler) reconcileWorkloadGroup(workloadGroup platformv1alpha1.Worklo
 
 	status := scheduledStatus
 	if work.IsResourceRequest() {
+		// If the Work is for a Resource Request, only one Workplacement will be created per
+		// WorkloadGroup. If this Workplacement already exists, it will be updated.
 		if len(existingWorkplacements) > 0 {
 			var errored int
 			for _, existingWorkplacement := range existingWorkplacements {
 				s.Log.Info("found workplacement for work; will try an update")
-				misscheduled, err := s.UpdateWorkPlacement(workloadGroup, work, &existingWorkplacement)
+				misscheduled, err := s.updateWorkPlacement(workloadGroup, work, &existingWorkplacement)
 				if err != nil {
 					s.Log.Error(err, "error updating workplacement for work", "workplacement", existingWorkplacement.Name, "work", work.Name, "workloadGroupID", workloadGroup.ID)
 					errored++
@@ -217,6 +197,7 @@ func (s *Scheduler) reconcileWorkloadGroup(workloadGroup platformv1alpha1.Worklo
 			if errored > 0 {
 				return "", fmt.Errorf("failed to update %d of %d workplacements for work", errored, len(existingWorkplacements))
 			}
+
 			return status, nil
 		}
 	}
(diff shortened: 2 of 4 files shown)


richcooper95:
All tests are passing after the migration, now just looking for general test improvements

controllers/destination_controller.go
```diff
@@ -92,7 +92,9 @@ func (r *DestinationReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return defaultRequeue, nil
 	}
 
-	if err := r.Scheduler.ReconcileDestination(); err != nil {
+	// TODO: Instead of this function call, we could watch for changes to the Destination
+	// CRD and triggering all Works to reconcile on every change.
```
kirederik, line 96: question: you mean, inside the scheduler?

if so, that's the end goal (to make the scheduler actually a controller)

controllers/scheduler.go
```diff
@@ -38,8 +38,11 @@ type Scheduler struct {
 	Log    logr.Logger
 }
 
-// Only reconciles Works that are from a Promise Dependency
-func (s *Scheduler) ReconcileDestination() error {
+// Reconciles all Works by scheduling each Work's WorkloadGroups to appropriate
+// Destinations.
+// Only reconciles Works that are from a Promise Dependency.
+// This is called by the Destination controller to schedule Workloads to Destinations.
```
kirederik, line 44: >// This is called by the Destination controller to schedule Workloads to Destinations.

is this comment (calling out who calls this function) necessary? it feels like one of those comments that can quickly get outdated and it's simple enough to find callers using any IDE.

controllers/scheduler.go
```diff
@@ -191,6 +170,12 @@ func (s *Scheduler) cleanupDanglingWorkplacements(work *platformv1alpha1.Work) e
 	return nil
 }
 
+// Reconciles a WorkloadGroup by scheduling it to a Destination via a Workplacement.
+// If the Work is for a Resource Request, only one Workplacement will be created. If
+// this Workplacement already exists, it will be updated.
+// If the Work is for a Promise Dependency, a Workplacement will be created for each
+// Destination that matches the WorkloadGroup's DestinationSelectors. If any of these
+// Workplacements already exist, they will be updated.
```
kirederik, line 178: would it be better to inline the comments at the point they happen and leave the top level comment as a general documentation? i fear this can quickly get out of date and never updated.

controllers/destination_controller.go
richcooper95: 👌 - the Scheduler watches Destinations and triggers Works to reconcile again on any change (I _think_ that makes sense!)

controllers/scheduler.go
richcooper95: Yep, you're right - will delete! (I initially put these comments in to help me understand the flow better)

controllers/scheduler.go
richcooper95: For resource Works, do we ever expect more than one WorkPlacement to be created for a WorkloadGroup (i.e. `len(existingWorkplacements) > 1`)?

I don't think we do, but we still iterate over them [below](https://github.com/syntasso/kratix/pull/59/files#diff-c70d9200060da1bcf5c33119592a7d74e3dab7b19cabf769686f0791372d8d0cL205)

Would it be better to assert that we only have one WorkPlacement before trying to update it?

kirederik: yeah, a resource can be now be scheduled across multiple destinations, so it may happen

controllers/scheduler_test.go
```diff
@@ -2,291 +2,518 @@ package controllers_test
...
+				Expect(workPlacements.Items[1].Namespace).To(Equal(KratixSystemNamespace))
+				Expect(workPlacements.Items[1].Spec.TargetDestinationName).To(Equal(devDestination2.Name))
+				Expect(workPlacements.Items[1].Spec.Workloads).To(Equal(dependencyWorkForDev.Spec.WorkloadGroups[0].Workloads))
+				Expect(workPlacements.Items[1].Spec.ID).To(Equal(dependencyWorkForDev.Spec.WorkloadGroups[0].ID))
+
+				Expect(workPlacements.Items[2].Name).To(HavePrefix("dev-work-name.dev-3"))
+				Expect(workPlacements.Items[2].Namespace).To(Equal(KratixSystemNamespace))
+				Expect(workPlacements.Items[2].Spec.TargetDestinationName).To(Equal(devDestination3.Name))
+				Expect(workPlacements.Items[2].Spec.Workloads).To(Equal(dependencyWorkForDev.Spec.WorkloadGroups[0].Workloads))
+				Expect(workPlacements.Items[2].Spec.ID).To(Equal(dependencyWorkForDev.Spec.WorkloadGroups[0].ID))
```
kirederik, line 98: (more fyi than not)

there are multiple tests where we check values on a struct

we could use [gstruct](https://onsi.github.io/gomega/#codegstructcode-testing-complex-data-types) from  Gomega and write something slightly more readable.

For example, see https://github.com/syntasso/kratix/blob/374ba26857874742c8c87ab0e3df110bf2f7a1ed/controllers/promise_and_dynamic_resource_request_controller_test.go#L900-L912

(if we dot import gstruct it can read even better)


This particular snippet may be rewritten as

```go
id := func(index int, _ interface{}) string {
	return strconv.Itoa(index)
}
Expect(workPlacements.Items).To(MatchElementsWithIndex(id, IgnoreExtras, Elements{
	"0": MatchFields(IgnoreExtras, Fields{
		"ObjectMeta": MatchFields(IgnoreExtras, Fields{
			"Name":      HavePrefix("dev-work-name.dev-1"),
			"Namespace": Equal(KratixSystemNamespace),
		}),
		"Spec": MatchFields(IgnoreExtras, Fields{
			"TargetDestinationName": Equal(devDestination.Name),
			"Workloads":             Equal(dependencyWorkForDev.Spec.WorkloadGroups[0].Workloads),
			"ID":                    Equal(dependencyWorkForDev.Spec.WorkloadGroups[0].ID),
		}),
	}),
}))
```

controllers/scheduler_test.go
```diff
@@ -2,291 +2,518 @@ package controllers_test
...
 
+				It("does not change the existing WorkPlacement", func() {
+					Expect(fakeK8sClient.List(context.Background(), &workPlacements)).To(Succeed())
 					Expect(workPlacements.Items).To(HaveLen(1))
+					newWorkPlacement := workPlacements.Items[0]
+
+					// compare each field in the work placements to ensure they are the same
+					Expect(newWorkPlacement.Name).To(Equal(workPlacement.Name))
+					Expect(newWorkPlacement.Namespace).To(Equal(workPlacement.Namespace))
+					Expect(newWorkPlacement.ObjectMeta.Labels["kratix.io/work"]).To(Equal(workPlacement.ObjectMeta.Labels["kratix.io/work"]))
```
kirederik, line 178: nit: very long lines... should we break?

controllers/destination_controller.go
richcooper95: In the future, we want to make the Scheduler a Controller, which watches both Destinations _and_ Works:
- if a new Destination is added, reconcile all Works
- if a new Work is added, reconcile _that_ Work

controllers/scheduler_test.go
```diff
@@ -2,291 +2,518 @@ package controllers_test
...
+
+					Expect(fakeK8sClient.List(context.Background(), &workPlacements)).To(Succeed())
+					Expect(workPlacements.Items).To(HaveLen(1))
+					workPlacement = workPlacements.Items[0]
+
+					_, err = scheduler.ReconcileWork(latestWork)
+					Expect(err).ToNot(HaveOccurred())
+				})
 
+				It("does not change the existing WorkPlacement", func() {
```
kirederik, line 170: can we capture the previous resource version and assert that it hasn't change?

controllers/scheduler_test.go
```diff
@@ -334,212 +563,334 @@ var _ = Describe("Controllers/Scheduler", func() {
...
 
 						devWorkPlacement := workPlacements.Items[0]
 						Expect(devWorkPlacement.Spec.TargetDestinationName).To(Equal(devDestination.Name))
 						Expect(devWorkPlacement.ObjectMeta.Labels["kratix.io/work"]).To(Equal(dependencyWorkForDev.Name))
+						Expect(devWorkPlacement.ObjectMeta.ResourceVersion < devWorkPlacementResourceVersionPreReconcile).To(BeTrue())
 
 						devWorkPlacement2 := workPlacements.Items[1]
 						Expect(devWorkPlacement2.Spec.TargetDestinationName).To(Equal(devDestination2.Name))
 						Expect(devWorkPlacement2.ObjectMeta.Labels["kratix.io/work"]).To(Equal(dependencyWorkForDev.Name))
+						Expect(devWorkPlacement2.ObjectMeta.ResourceVersion > devWorkPlacement2ResourceVersionPreReconcile).To(BeTrue())
```
kirederik, line 709: tip: [BeNumerically](https://onsi.github.io/gomega/#benumericallycomparator-string-compareto-interface)

kirederik:
```
Expect(num).To(BeNumerically(">", 2))
```

controllers/scheduler_test.go
```diff
@@ -334,212 +563,334 @@ var _ = Describe("Controllers/Scheduler", func() {
...
-						Expect(devWorkPlacement.ObjectMeta.Labels["kratix.io/work"]).To(Equal(dependencyWorkForDev.Name))
+							This means that we need to assert that the latest ResourceVersion is
+							_greater than_ the pre-reconcile ResourceVersion, but it feels like
+							we should be able to assert that the ResourceVersion is the _same_
+							(i.e. that devWorkPlacement2 doesn't get touched).
 
-						devWorkPlacement2 = workPlacements.Items[1]
-						Expect(devWorkPlacement2.Spec.TargetDestinationName).To(Equal(devDestination2.Name))
-						Expect(devWorkPlacement2.ObjectMeta.Labels["kratix.io/work"]).To(Equal(dependencyWorkForDev.Name))
+							Is there a better way to assert that devWorkPlacement2 is not recreated?
```
kirederik, line 729: you can use `creationTimestamp` instead; set it manually if its not created automatically by the fakeClient (or in the controller directly)

controllers/scheduler_test.go
kirederik: alternative:

```
Expect(newWorkPlacement.Labels["kratix.io/work"]).To(Equal(workPlacement.Labels["kratix.io/work"]))
```

kirederik: if all labels match:
```go
Expect(newWorkPlacements.Labels).To(Equal(workPlacement.Labels["kratix.io/work"]))
```

controllers/scheduler.go
richcooper95: Just recapping what we just talked about:
- multi-destination scheduling allows a single `Work` to split across multiple `Destination`s
- it does this by defining a set of `WorkloadGroups` in a `Work`, each of which is scheduled to a specific (or all) `Destination`s (via `DestinationSelectors`)
- the `Scheduler` creates a _single_ `WorkPlacement` for each `WorkloadGroup` in order to place it in the given `Destination` (i.e. we _do_ only expect max. 1 `WorkPlacement` to be returned above)
- for dependency `Work`s, each `WorkloadGroup` is scheduled to _all_ matching `Destination`s
- for resource `Work`s, each `WorkloadGroup` is scheduled to _only one_ matching `Destination`

Shout if I've not quite covered anything right! 😊

richcooper95:
@kirederik -- with research day tomorrow and then Christmas, I'm planning to leave the `gstruct` and line length stuff for now (with a chore put on the backlog to look at introducing `gstruct`) if that's alright with you!

controllers/scheduler_test.go
richcooper95: Created [186726492](https://www.pivotaltracker.com/story/show/186726492) to start using gstruct ✅

controllers/scheduler_test.go
richcooper95: Created [186726492](https://www.pivotaltracker.com/story/show/186726492) to start using gstruct ✅

controllers/scheduler_test.go
richcooper95: > CreationTimestamp is a timestamp representing the server time when this object was created. It is not guaranteed to be set in happens-before order across separate operations. Clients may not set this value. It is represented in RFC3339 form and is in UTC. Populated by the system. Read-only. Null for lists.

Looks like we can't set `CreationTimestamp` 😢

richcooper95: And because the fake client doesn't set it, I don't think we can use it to compare in the tests

richcooper95: To get this PR out, I'm just going to assert that `devWorkPlacement2` is present for now, and add a TODO to ensure it's the same object!

richcooper95: For posterity, here are the reasons why this is harder to test than it should be:
- the CreationTimestamp is not set by the fake client
- the UID is not set by the fake client
- the ResourceVersion _is_ set, but is updated on every reconcile because the status is updated* (we could assert that it's a known value here, but it might change if the code under test calls out to update the status any more than it currently does, so this might be flaky)

*This might be fake-client-specific - there are a couple of similar-ish Github issues, but I'm stopping investigation here! I did find out that it's definitely the status update in the Scheduler that's bumping the resource version, not the pointer-to-true candidate I thought about in a previous comment

richcooper95: Also, we might not even _want_ to test this, since all we really care about (I think) is that the WorkPlacement for `devWorkPlacement2` exists, regardless of whether we recreate it or not

PR #65, opened 2023-12-28
Minimal git system test

kirederik:


.circleci/config.yml
```diff
@@ -142,24 +142,18 @@ jobs:
...
-    executor: machine-large
-    steps:
-      - install_software
+            make test
       - run:
           name: System tests
           command: |
-            make test
-            SYSTEM_TEST_STORE_TYPE=bucket DOCKER_BUILDKIT=1 ACK_GINKGO_RC=true make system-test
+            DOCKER_BUILDKIT=1 ACK_GINKGO_RC=true make system-test
```
aclevername, line 156: better

test/system/system_test.go
```diff
@@ -580,8 +552,113 @@ var _ = Describe("Kratix", func() {
...
+					set -x
+					if [ "${KRATIX_WORKFLOW_ACTION}" = "configure" ]; then :
+						echo "message: My awesome status message" > /kratix/metadata/status.yaml
+						echo "key: value" >> /kratix/metadata/status.yaml
+						mkdir -p /kratix/output/foo/
+						echo "{}" > /kratix/output/foo/example.json
+						kubectl get namespace imperative-$(yq '.metadata.name' /kratix/input/object.yaml) || kubectl create namespace imperative-$(yq '.metadata.name' /kratix/input/object.yaml)
+						exit 0
+					fi
+					kubectl delete namespace imperative-$(yq '.metadata.name' /kratix/input/object.yaml)
```
aclevername, line 648: better as an asset file?

aclevername (approved):
lgtm

test/system/system_test.go
kirederik: its one of those... i think this was an asset file before, then it became in-line... 🤔
(i just copied over from other test into a function so i could re-use)

aclevername: fair

PR #66, opened 2023-12-29
Improve PromiseRelease Status and Conditions

aclevername:


api/v1alpha1/promiserelease_webhook.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	promise, err := promiseFetcher.FromURL(r.Spec.SourceRef.URL)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch promise: %w", err)
+	}
+
+	promiseVersion, found := promise.GetLabels()[PromiseVersionLabel]
+	if !found {
+		msg := fmt.Sprintf("Warning: version label (%s) not found on promise, installation will fail", PromiseVersionLabel)
```
kirederik, line 62: why not error then?

api/v1alpha1/promiserelease_webhook.go
```diff
@@ -0,0 +1,99 @@
...
+
+	if promiseVersion != r.Spec.Version {
+		msg := fmt.Sprintf("Warning: version labels do not match, found: %s, expected: %s, installation will fail", promiseVersion, r.Spec.Version)
+		return []string{msg}, nil
+	}
+
+	return nil, nil
+}
+
+func (r *PromiseRelease) ValidateUpdate(old runtime.Object) (admission.Warnings, error) {
```
kirederik, line 74: it looks like we're not pulling and running the same checks as we do in `create`. curious to know why not

kirederik: i see the comment below on the test; there are ways to avoid doing a re-fetch on updates though 🤔

api/v1alpha1/promiserelease_webhook.go
```diff
@@ -0,0 +1,99 @@
...
+
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+const (
+	KratixPrefix        = "kratix.io/"
```
kirederik, line 30: a constant with the same value is also declared in controllers/utils.go. we should converge to a single definition.

controllers/promiserelease_controller.go
```diff
@@ -84,35 +88,44 @@ func (r *PromiseReleaseReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 		return ctrl.Result{}, fmt.Errorf("failed to create or update promise: %w", err)
 	}
 
-	promiseRelease.Status.Installed = true
-	if err := r.Client.Status().Update(ctx, promiseRelease); err != nil {
-		return ctrl.Result{}, fmt.Errorf("failed to update PromiseRelease status: %w", err)
-	}
-
+	promiseRelease.Status.Status = statusInstalled
+	r.updateStatus(opts, promiseRelease, statusInstalled, conditionMessageInstalled, conditionReasonInstalled)
```
kirederik, line 128: any particular reason why we're ignoring all the errors when updating status?

controllers/promiserelease_controller.go
```diff
@@ -196,43 +209,74 @@ func (r *PromiseReleaseReconciler) delete(o opts, promiseRelease *v1alpha1.Promi
...
-		err = o.client.Status().Update(o.ctx, promiseRelease)
-		if err != nil {
-			return ctrl.Result{}, fmt.Errorf("failed to update PromiseRelease status: %w", err)
+		return false, nil
+	default:
+		return false, fmt.Errorf("expected 0 or 1 promises, got %d", len(promises.Items))
+	}
+}
+
+func (r *PromiseReleaseReconciler) updateStatus(o opts, pr *v1alpha1.PromiseRelease,
```
kirederik, line 235: it looks like this method is updating the Installed condition + the object status.status; name could be a bit more descriptive

controllers/promiserelease_controller.go
```diff
@@ -196,43 +209,74 @@ func (r *PromiseReleaseReconciler) delete(o opts, promiseRelease *v1alpha1.Promi
...
+		Reason:             conditionReason,
+		Status:             conditionStatus,
+		LastTransitionTime: v1.Time{Time: time.Now()},
+	}
+
+	if existingConditionIndex == -1 {
+		pr.Status.Conditions = append(pr.Status.Conditions, condition)
+	} else {
+		pr.Status.Conditions[existingConditionIndex] = condition
+	}
```
kirederik, line 267:
```suggestion
	condition := v1.Condition{
		Type:               "Installed",
		Message:            conditionMessage,
		Reason:             conditionReason,
		Status:             conditionStatus,
	}

        meta.SetStatusCondition(&pr.Status.Conditions, condition)
```

controllers/promiserelease_controller.go
```diff
@@ -196,43 +209,74 @@ func (r *PromiseReleaseReconciler) delete(o opts, promiseRelease *v1alpha1.Promi
...
+	for i, condition := range pr.Status.Conditions {
+		if condition.Type == "Installed" {
+			existingConditionIndex = i
+			if condition.Message == conditionMessage &&
+				condition.Reason == conditionReason {
+				//don't update the status if its already correct
+				return
+			}
 		}
+	}
```
kirederik, line 248:
```suggestion
	existingCondition := meta.FindStatusCondition(pr.Status.Conditions, "Installed")
	if existingCondition != nil {
		if existingCondition.Message == conditionMessage && existingCondition.Reason == conditionReason {
			//don't update the status if its already correct
			return
		}
	}
```

api/v1alpha1/promiserelease_webhook.go
kirederik: i can also see in other places we have replaced the controllers constant with this constant, but not in all 🤔

controllers/promiserelease_controller_test.go
```diff
@@ -92,79 +94,105 @@ var _ = Describe("PromiseReleaseController", func() {
...
-				It("installs the promise from the URL", func() {
-					_, err := t.reconcileUntilCompletion(reconciler, &promiseRelease)
+				It("errors", func() {
 					Expect(err).To(MatchError("failed to fetch promise from url: can't do mate"))
 				})
+
+				It("updates the status", func() {
+					err := fakeK8sClient.Get(context.TODO(), promiseReleaseNamespacedName, &promiseRelease)
+					Expect(err).NotTo(HaveOccurred())
+					Expect(promiseRelease.Status.Status).To(Equal("Error installing"))
```
kirederik, line 123: status could be a bit more descriptive then always error 🤔

controllers/promiserelease_controller.go
```diff
@@ -196,43 +209,74 @@ func (r *PromiseReleaseReconciler) delete(o opts, promiseRelease *v1alpha1.Promi
...
 	}
 
 	switch len(promises.Items) {
 	case 1:
-		if !promises.Items[0].GetDeletionTimestamp().IsZero() {
-			return defaultRequeue, nil
-		}
-		if promises.Items[0].Labels[promiseVersionLabel] == promiseRelease.Spec.Version {
-			break
+		if promises.Items[0].Labels[v1alpha1.PromiseVersionLabel] == promiseRelease.Spec.Version {
```
kirederik, line 224:
```suggestion
		return promises.Items[0].Labels[v1alpha1.PromiseVersionLabel] == promiseRelease.Spec.Version, nil
```

kirederik: ^ the above is not commit-able -- gotta change the entire block

api/v1alpha1/promiserelease_webhook.go
aclevername: I think its a balance with webhooks. The state of the world not being okay when you declare a resource is sort of fine? you can resolve the external stuff afterwards and the system will reconcile. I think this whole PR has been a balance of this.

api/v1alpha1/promiserelease_webhook.go
aclevername: How? I think `update` events could happen very frequently (e.g. flux applying every 5 seconds, so keeping the api call low-cost/footprint is good

api/v1alpha1/promiserelease_webhook.go
aclevername: thought I got them all! must of been a merge miss, I'll update the remaining!

controllers/promiserelease_controller.go
aclevername: nice

api/v1alpha1/promiserelease_webhook.go
kirederik: we could, for example, base64 the url and add as a label in the promiserelease at create, and check if the url has changed on update

it wouldn't cover the case of when the remote itself has changed but that's fine

controllers/promiserelease_controller_test.go
kirederik: yeah, fair enough; i think `Error Installing` kinda annoys me more than if it was just `Error` lol but nw

kirederik (approved):

PR #69, opened 2024-01-11
Fix delete pipelines

richcooper95:
#67 introduced a bug which meant that the ServiceAccount used in delete pipelines wasn't configured correctly, which meant deletes failed.

tl;dr - we started to use the `Kind` and not the plural for the Role resource, and didn't catch it in tests because `bash` and `redis` both have identical singular and plural versions 😢

This PR fixes the bug by passing in the plural when creating the delete pipeline, and also makes `bashes` the plural of `bash` so we have some coverage of this case in the system tests.

(There's also a `redis` promise used in the system tests, so we still have some coverage of the identical-singular-plural case.)

See [here](https://syntasso.slack.com/archives/C044T9ZFUMN/p1704970369264829?thread_ts=1704900358.243089&cid=C044T9ZFUMN) in Slack for more detail 🧵

### DIFF
--- a/controllers/util.go
+++ b/controllers/util.go
@@ -159,18 +159,12 @@ func ensureDeletePipelineIsReconciled(jobOpts jobOpts) (ctrl.Result, error) {
 	}
 
 	if existingDeletePipeline == nil {
+		jobOpts.logger.Info("Creating Delete Pipeline. The pipeline will now execute...")
+
+		//TODO retrieve error information from applyResources to return to the caller
 		applyResources(jobOpts.opts, jobOpts.pipelineResources...)
+
 		return defaultRequeue, nil
-		// jobOpts.logger.Info("Creating Delete Pipeline. The pipeline will now execute...")
-		// //TODO come back and address- do we know for sure that the cluster roles and service accounts will exist?
-		// err = jobOpts.client.Create(jobOpts.ctx, jobOpts.pipelineResources[0])
-		// if err != nil {
-		// 	jobOpts.logger.Error(err, "Error creating delete pipeline")
-		// 	y, _ := yaml.Marshal(&jobOpts.pipelineResources[0])
-		// 	jobOpts.logger.Error(err, string(y))
-		// 	return ctrl.Result{}, err
-		// }
-		// return defaultRequeue, nil
 	}
 
 	jobOpts.logger.Info("Checking status of Delete Pipeline")
@@ -180,9 +174,9 @@ func ensureDeletePipelineIsReconciled(jobOpts jobOpts) (ctrl.Result, error) {
 		if err := jobOpts.client.Update(jobOpts.ctx, jobOpts.obj); err != nil {
 			return ctrl.Result{}, err
 		}
+	} else {
+		jobOpts.logger.Info("Delete Pipeline not finished", "status", existingDeletePipeline.Status)
 	}
-
-	jobOpts.logger.Info("Delete Pipeline not finished", "status", existingDeletePipeline.Status)
 
 	return fastRequeue, nil
 }
@@ -327,6 +321,7 @@ func getJobsWithLabels(o opts, jobLabels map[string]string, namespace string) ([
 	return jobs.Items, nil
 }
 
+// TODO return error info (summary of errors from resources?) to the caller, instead of just logging
 func applyResources(o opts, resources ...client.Object) {
 	o.logger.Info("Reconciling pipeline resources")
 
--- a/lib/pipeline/assets/promise.yaml
+++ b/lib/pipeline/assets/promise.yaml
@@ -1,7 +1,7 @@
 apiVersion: platform.kratix.io/v1alpha1
 kind: Promise
 metadata:
-  name: redis
+  name: custom-namespace
   labels:
     kratix.io/promise-version: v1.2.0
     clashing-label: "new-promise-v2-value"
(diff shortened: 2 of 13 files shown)


controllers/util.go
```diff
@@ -159,18 +159,12 @@ func ensureDeletePipelineIsReconciled(jobOpts jobOpts) (ctrl.Result, error) {
...
 
 	if existingDeletePipeline == nil {
+		jobOpts.logger.Info("Creating Delete Pipeline. The pipeline will now execute...")
+
+		//TODO retrieve error information from applyResources to return to the caller
 		applyResources(jobOpts.opts, jobOpts.pipelineResources...)
+
 		return defaultRequeue, nil
-		// jobOpts.logger.Info("Creating Delete Pipeline. The pipeline will now execute...")
-		// //TODO come back and address- do we know for sure that the cluster roles and service accounts will exist?
```
abangser, line 165: These todo's are pretty different in concept. Is there a reason we have moved from needing to verify the roles vs needing to retrieve error info?

abangser, line 165: These todos (the one you removed vs the one you added) are pretty different in concept. Is there a reason we have moved from needing to verify the roles vs needing to retrieve error info?

controllers/util.go
```diff
@@ -180,10 +174,10 @@ func ensureDeletePipelineIsReconciled(jobOpts jobOpts) (ctrl.Result, error) {
 		if err := jobOpts.client.Update(jobOpts.ctx, jobOpts.obj); err != nil {
 			return ctrl.Result{}, err
 		}
+	} else {
+		jobOpts.logger.Info("Delete Pipeline not finished", "status", existingDeletePipeline.Status)
```
abangser, line 178: 💅 This is likely polish, but would it be helpful to also log an info on success? I ask because if delete succeeds the k8s resources disappear. It may be nice to have seen it in the logs 💭

lib/pipeline/delete.go
```diff
@@ -33,7 +33,7 @@ func newDelete(obj *unstructured.Unstructured, pipelines []platformv1alpha1.Pipe
 
 	resources := []client.Object{
 		serviceAccount(args),
-		role(obj, obj.GetKind(), args),
+		role(obj, objPlural, args),
```
aclevername, line 36: ahhh 🙈

aclevername (approved):
lgtm, 1 comment

lib/pipeline/assets/promise.yaml
```diff
@@ -14,13 +14,13 @@ spec:
...
+      name: custom-namespaces.marketplace.kratix.io
     spec:
       group: marketplace.kratix.io
       names:
-        kind: redis
-        plural: redis
-        singular: redis
+        kind: custom-namespace
+        plural: custom-namespaces
+        singular: custom-namespace
```
abangser, line 23: Am I correct that this will mean that we have both (this and bash/bashes) our test data now having different singular and plurals?

I would prefer to see us cover both scenarios. That being said, it would be pretty cryptic that we were doing that, so it may require some decent descriptions in either the variable names or comments!

abangser (commented):
Looks sound, submitting now so you can look at the comments that I have made and will do a test locally in the mean time. Thanks for this!

controllers/util.go
richcooper95: Ah yep - this TODO was addressed by the changes in #67 (we now ensure the roles/service accounts exist by creating them in the pipeline resources in `delete.go`)!

controllers/util.go
richcooper95: Do you mean we should log if the delete pipeline completed successfully? We had this already (a few lines above):

`jobOpts.logger.Info("Delete Pipeline Completed")`

lib/pipeline/assets/promise.yaml
richcooper95: These assets are only used in the pipeline unit tests, so I think just having the plural here is fine (since that's the field we want to have passed through) - we should definitely cover both scenarios in the system tests though!

In the system tests, we have some coverage of both:
- Different plural with `bash`/`bashes`
- Identical plural with `redis`/`redis` in the promise-with-requirement tests ([here](https://github.com/syntasso/kratix/blob/main/test/system/system_test.go#L145))

We could probably still make the system test coverage better (i.e. I'm not sure we cover all of the same assertions in each scenario) but that might be one for a separate PR!

lib/pipeline/delete.go
richcooper95: 🤦

lib/pipeline/delete.go
```diff
@@ -12,15 +12,15 @@ import (
...
 
-func NewDeleteResource(rr *unstructured.Unstructured, pipelines []platformv1alpha1.Pipeline, resourceRequestIdentifier, promiseIdentifier string) []client.Object {
-	return newDelete(rr, pipelines, resourceRequestIdentifier, promiseIdentifier)
+func NewDeleteResource(rr *unstructured.Unstructured, pipelines []platformv1alpha1.Pipeline, resourceRequestIdentifier, promiseIdentifier, crdPlural string) []client.Object {
+	return newDelete(rr, pipelines, resourceRequestIdentifier, promiseIdentifier, crdPlural)
 }
 
 func NewDeletePromise(promise *unstructured.Unstructured, pipelines []platformv1alpha1.Pipeline) []client.Object {
-	return newDelete(promise, pipelines, "", promise.GetName())
+	return newDelete(promise, pipelines, "", promise.GetName(), "promises")
```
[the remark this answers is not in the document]
richcooper95: It does now! https://github.com/syntasso/kratix/pull/69/commits/a2ce39ab3f952d85167034d9f21145a228722c63

PR #71, opened 2024-01-15
feat: add strict labels match to destinations

kirederik:
this property allows destinations to configure whether resources with no labels should be schedulable to them.

this is specially useful when registering non-k8s destinations;

for example, prior to this change:

if a destination for backstage is registered with a label `env: backstage`, every other destination must include a label (like `env: k8s`) and every promise must contain `destinationSelectors: { env: k8s }` just so that Kratix doesn't try to schedule k8s resources to the backstage destination.

with this change, all the user needs to do is set `strictLabelsMatch: true` on the backstage destination; with that property, kratix will only schedule to `backstage` works with the proper destinationSelector.


controllers/scheduler.go
```diff
@@ -441,11 +441,23 @@ func (s *Scheduler) getDestinationsForWorkloadGroup(destinationSelectors map[str
...
 	if err != nil {
 		s.Log.Error(err, "Error listing available Destinations")
 	}
-	return destinations.Items
+
+	if len(destinationSelectors) > 0 {
+		return destinationList.Items
+	}
+
+	destinations := []platformv1alpha1.Destination{}
```
SaphMB, line 453: Just to confirm my understanding:
- Should a user set `StrictMatchLabels: true` on a destination, this destination will not be returned as a potential destination by `getDestinationsForWorkloadGroup`
- Should the user want to schedule work to this destination, it would have been specified via a `DestinationSelector` and `getDestinationsForWorkloadGroup` would have returned before even checking whether the `StrictMatchLabels: true` existed on any destinations

api/v1alpha1/destination_types.go
```diff
@@ -49,6 +49,15 @@ type DestinationSpec struct {
...
 	StateStoreRef        *StateStoreReference `json:"stateStoreRef,omitempty"`
+
+	// By default, Kratix will schedule works without labels to all destinations
+	// (for promise dependencies) or to a random destination (for resource
+	// requests). If StrictMatchLabels is true, Kratix will only schedule works
+	// to this destination if the labels match strictly (that is, an empty label
+	// set on the work won't be schedule to this destination, unless the
+	// destination label set is also empty).
+	// +kubebuilder:validation:Optional
+	StrictMatchLabels bool `json:"strictMatchLabels,omitempty"`
```
aclevername, line 60: > that is, an empty label set on the work won't be schedule to this destination, unless the destination label set is also empty

This is the only use case right? I think its hard to drive that understanding from the key name, I'd definitely think it means "only scheduling Work to this that have all my Destinations labels, not a subset". But I cannot think of a better name to describe what the comment highlights well. I think its a smell this behaviour from Kratix is weird... but as discussed yesterday we aren't going to change it yet. So I guess :lgtm: 🤷

kirederik: yeah, correct.

this is really "dontScheduleToThisDestinationsIfNoLabelsAreSet".

We _could_ make it _really_ strict matching, as in, only schedule to this destination if all `labels match` rather than a subset 🤔

controllers/scheduler.go
kirederik: hm... not sure i understood your understanding 😆

> Should a user set `StrictMatchLabels: true `on a destination, this destination will not be returned as a potential destination by `getDestinationsForWorkloadGroup`

It could be returned if the labels match 🤔

> Should the user want to schedule work to this destination, it would have been specified via a `DestinationSelector` and `getDestinationsForWorkloadGroup` would have returned before even checking whether the `StrictMatchLabels: true` existed on any destinations

yeah
basically if there are `destinationSelectors` and `Destination A` was selected, it means the labels on A match the `destinationSelectors` so no need to check if "strict" is set to true.

SaphMB: Exactly this, I inverted the logic as I started with looking and this snippet and then validated how that destination _could_ have been chosen if desired (which is part of the logic earlier in the function).

api/v1alpha1/destination_types.go
SaphMB: This comment is a little confusing, could it be clearer to talk about this in relation to only matching the DestinationSelectors since they are the only labels in question. Something like `matchDestinationSelector` could be fitting as we aren't being so strict as to match all of them.

Conversely, this could exist in the comment itself e.g. 

```
	// By default, Kratix will schedule works without labels to all destinations
	// (for promise dependencies) or to a random destination (for resource
	// requests). If StrictMatchLabels is true, Kratix will only schedule works
	// to this destination if destination matches the labels - reflecting the 
        // DestinationSelector configured on the Promise. An empty label
	// set on the work won't be schedule to this destination, unless the
	// destination label set is also empty
```

kirederik: @SaphMB are you suggesting we change the name of the field? From `StrictMatchLabels` to `MatchDestinationSelector`?

(i like the comment btw, will update)

SaphMB: I'm very wary that there is a difference between what is clearer for me and what is clearer for the majority of users. That's what I was suggesting if that would actually be more useful for users who aren't me, but I think the updated comment would also suffice

On that note, I did make some typos there:
```
	// By default, Kratix will schedule works without labels to all destinations
	// (for promise dependencies) or to a random destination (for resource
	// requests). If StrictMatchLabels is true, Kratix will only schedule works
	// to this destination if the destination matches the labels - reflecting the 
        // DestinationSelector configured on the Promise. An empty label
	// set on the work won't be scheduled to this destination, unless the
	// destination label set is also empty
```

kirederik: I think I like `MatchDestinationSelectors` better tbf -- i was just double checking my understanding 😛

kirederik: I'll change the comment-documentation to say 

> If StrictMatchLabels is true, Kratix will only schedule works to this destination if the destination can be matched by the defined DestinationSelectors

PR #79, opened 2024-02-07
pipeline-adapter kubectl match target platform

deltaroe:


abangser:
Thanks so much @deltaroe for pointing this out. We were relying on the rosetta emulations on Mac which [had a regression](https://github.com/docker/for-mac/issues/7180#issuecomment-1943274145) recently. While the regression appears fixed in 14.3 we also want to be resilient to this issue in the future so have solved this in a slightly different way [here](https://github.com/syntasso/kratix/pull/86/files#diff-398b008c6df4123ba627b607f60fcd3e0b921982a5c945c03f1a32198d7c14beR30) and tested in on a few MacOS Silicon machines at different versions.

I will close this for now, but please do shout if you (or anyone else!) see a problem again.

Issue #95, opened 2024-03-19
bug: Specifying additional pod properties (e.g. Command) isnt supported for delete workflows, but the API accepts it


aclevername:
```
apiVersion: platform.kratix.io/v1alpha1
kind: Promise
metadata:
  name: app
spec:
  api: # ...
  dependencies: # ...
  workflows:
    resource:
      configure:
        - apiVersion: platform.kratix.io/v1alpha1
          kind: Pipeline
          metadata:
            name: resource-configure
          spec:
            containers:
              - name: create-resources
                image: kratix-workshop/app-pipeline-image:v1.0.0
                command: [ resource-configure ]
      delete:
        - apiVersion: platform.kratix.io/v1alpha1
          kind: Pipeline
          metadata:
            name: resource-configure
          spec:
            containers:
              - name: create-resources
                image: kratix-workshop/app-pipeline-image:v1.0.0
                command: [ resource-delete ]
```

when the pipeline runs for configure, `resource-configure` is provided as the `command` for the pod. But for delete, nothing is specified. The same is true for a bunch of other variables, e.g. `envFrom`. The full list can be seen below.

looks like in the code here we pass it through for configure:
https://github.com/syntasso/kratix/blob/c964c11111c272da030c090df3234f273ac28903/lib/pipeline/configure.go#L181-L189

but we don't for delete:
https://github.com/syntasso/kratix/blob/c964c11111c272da030c090df3234f273ac28903/lib/pipeline/delete.go#L80-L90

SaphMB:
Fixes #96 

The functionality for configuring Jobs for Pipelines now sits in a central place such that any future updates to the properties of Containers will applied to both Delete and Configure pipelines

SaphMB:
Fixes #96 

The functionality for configuring Jobs for Pipelines now sits in a central place such that any future updates to the properties of Containers will applied to both Delete and Configure pipelines

PR #96, opened 2024-03-19
feat: ensure all delete pipeline params are passed to pods

richcooper95:
This commit ensures that we pass through all the fields from the Kratix Pipeline object to the Pods that are created for a delete Job.


SaphMB:
Fixes #95 

The functionality for configuring Jobs for Pipelines now sits in a central place such that any future updates to the properties of Containers will applied to both Delete and Configure pipelines

PR #107, opened 2024-04-26
Small fixes

ChunyiLyu:
Commit messages should be self explanatory. Most of them are picked up by my IDE.
Have run unit and system tests locally.


controllers/promise_controller.go
```diff
@@ -118,14 +118,14 @@ func (r *PromiseReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
-			return ctrl.Result{}, nil
-		}
+
+	if client.IgnoreNotFound(err) != nil {
 		r.Log.Error(err, "Failed getting Promise", "namespacedName", req.NamespacedName)
 		return defaultRequeue, nil
+	} else if errors.IsNotFound(err) {
+		return ctrl.Result{}, nil
 	}
+
```
kirederik, line 128: i like this, but i found the if/else/if a bit confusing :p 

maybe dropping the else?

```go
if errors.IsNotFound(err) {
  return ctrl.Result{}, nil
}
if client.IgnoreNotFound(err) != nil {
  r.Log.Error(err, "Failed getting Promise", "namespacedName", req.NamespacedName)
  return defaultRequeue, nil
}
```

ChunyiLyu:
@kirederik thanks for the review. I'm going to close this one and recreate because it seems I should create it against dev.

controllers/promise_controller.go
ChunyiLyu: Thanks Derik. Will include in the new PR

PR #123, opened 2024-05-13
Remove dev branch as gate to main

abangser:
Originally we used the dev branch as a way to mitigate risk against breaking users with changes. This was before we had introduced the release process which is itself a safety gate for this.

This change removes the dev branch from the CI process, enabling engineers to open PRs directly against main branch. 

Note that this does not change the GitHub settings so main continues to be the "default" branch:
image

And we will continue to have branch protection rules on `main` which means only Kratix maintainers can push directly to that branch.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -449,7 +449,7 @@ workflows:
           cron: "0 0 * * 1"
           filters:
             branches:
-              only: [dev]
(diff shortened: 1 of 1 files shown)


abangser:
Existing main protections that I don't expect to change:
image

PR #128, opened 2024-05-16
feat: respect promise api AdditionalPrinterColumns

kirederik:
this allows promise writers to control what kubectl shows when `kubectl get <promise crd>` is executed on the terminal

### DIFF
--- a/controllers/promise_controller.go
+++ b/controllers/promise_controller.go
@@ -938,12 +938,14 @@ func setStatusFieldsOnCRD(rrCRD *apiextensionsv1.CustomResourceDefinition) {
 			Status: &apiextensionsv1.CustomResourceSubresourceStatus{},
 		}
 
-		rrCRD.Spec.Versions[i].AdditionalPrinterColumns = []apiextensionsv1.CustomResourceColumnDefinition{
-			{
-				Name:     "status",
-				Type:     "string",
-				JSONPath: ".status.message",
-			},
+		if len(rrCRD.Spec.Versions[i].AdditionalPrinterColumns) == 0 {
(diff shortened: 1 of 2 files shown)


controllers/promise_controller.go
```diff
@@ -938,12 +938,14 @@ func setStatusFieldsOnCRD(rrCRD *apiextensionsv1.CustomResourceDefinition) {
 			Status: &apiextensionsv1.CustomResourceSubresourceStatus{},
 		}
 
-		rrCRD.Spec.Versions[i].AdditionalPrinterColumns = []apiextensionsv1.CustomResourceColumnDefinition{
-			{
-				Name:     "status",
-				Type:     "string",
-				JSONPath: ".status.message",
-			},
+		if len(rrCRD.Spec.Versions[i].AdditionalPrinterColumns) == 0 {
```
abangser, line 941: Is there an expected behaviour here about merging things? As in, if the user does not set status in their additional columns, should we keep status? I think this was the only other thing discussed a while back.

kirederik: > caveat: i don't remember being in a conversation about this

i thought about merging, but decided against it. my reasoning: 
* not all users will set the columns; for those, adding the `.status.message` is the standard Kratix behaviour
* the users that define the additional columns should have _full_ control of what's displayed
  * that means, if we merge by default, i'd say we would need to provide a way for it to be hidden 
  * with the replace approach, they can always add the `status.message` column by adding it themselves to the list

kirederik: merging as is for now, we can always change it if users find it too confusing

PR #130, opened 2024-05-17
feat: Add support for writing to top-level dir of statestore

### LINKED ISSUE #127: As Penny I want to be able to write files to the top-level directory of a destination instead of in nested directories
## Background
Today when files are written to a destination they are nested in a directory structure. The structure takes the following format for resources (and a slightly different for promise, see [here](https://docs.kratix.io/main/reference/destinations/intro))
```
statestore.Spec.Path/
    destination.Spec.Path/
        destination.Name/
            resources/
                resource.Namespace/
                    promise.Name/
                        resource.Namespace/
```

What this might look like in practise for a `Redis` promise where an`example` request has been made and outputs the following to `/kratix/output/`
```
redis-instance.yaml
subdir/
  redis-secret.yaml
```
```
kratix/platform-cluster/resources/default/redis/example/instance-configure/5058f/redis-instance.yaml
kratix/platform-cluster/resources/default/redis/example/instance-configure/5058f/subdir/redis-secret.yaml

```

`5058f` is the 5-char [md5sum](https://github.com/syntasso/kratix/blob/5f16b20490730e4f1bf2377614cd73f3e229cdbf/work-creator/pipeline/work_creator.go#L68) is used to differniate the different schedling "groups" from within the pipeline. 5058f is the default group (its the first 5 characters of the md5sum of ".")

While this approach works well for many different types of destinations, those that cannot handle nested directories or have strict rules about file names/directories cannot work with this. We can get around this by making the usage of nested directories optional.

## Proposal

Enable `Destinations` to configure off the usage of nested directories via the spec. Up the implementors to decide on the correct format, but something like:

```
spec:
    filepathExpression:
      type: nested | flat # nested is default
```

We should ensure **the default behaviour remains as is, so users have to opt in to this change**.

If a destination was configured to use this new method, and the above Redis request was scheduled to it you'd expect the following to be written:
```
redis-instance.yaml
subdir/redis-secret.yaml
```

This opens the door to naming clashes, which will *have to be documented*.  In #129 we will introduce support for specifying `.filepathExpression.promise` and `.filepathExpression.resource`, to allow full control

aclevername:
closes #127, paired with @abangser


aclevername:
Waiting for tests to pass before merging

PR #136, opened 2024-05-22
fix: rename json key filepath.type to filepath.mode

ChunyiLyu:
struct field name was renamed but we missed the json key name from previous PR


aclevername (approved):

aclevername:
whoooops

PR #150, opened 2024-05-30
bugfix: use unique name for static dependencies

### LINKED ISSUE #149: [Bug] Installing multiple promise with static dependencies into a destination that has .spec.filepath.mode=none breaks the installed promises
If. a promise has statically defined dpendencies, e.g https://github.com/syntasso/kratix-marketplace/blob/c1b69be8b30a397c7722c5065216832e6ac37dd3/vault-namespace/promise.yaml#L45
```
kind: Promise
metadata:
  name: vaultnamespace
  namespace: default
spec:
  api:
  ...
  dependencies:
    - apiVersion: rbac.authorization.k8s.io/v1
      kind: Role
      metadata:
        name: vault-promise
      ...
```

this results in a `static/dependencies.yaml` file getting written to the statestore. In our default `nestedByMetadata` world this is fine, as this is all separated under different directories. When we are having all promises write to the top level directory this results in mulitple promises all trying to write the same file with the same name.

aclevername:
Closes #149

### DIFF
--- a/api/v1alpha1/work_types.go
+++ b/api/v1alpha1/work_types.go
@@ -17,6 +17,8 @@ limitations under the License.
 package v1alpha1
 
 import (
+	"fmt"
+
 	"github.com/syntasso/kratix/lib/hash"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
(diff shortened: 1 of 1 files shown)


ChunyiLyu (approved):
LGTM

PR #151, opened 2024-05-30
feat: surface writer metadata as status in workplacement

kirederik:
- **refactor: merge writer funcs into a single one**
- **feat: surface version to the workplacement status**
- **feat: attempt to preserve version id on failures**

### DIFF
--- a/controllers/suite_test.go
+++ b/controllers/suite_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package controllers_test
 
 import (
+	"context"
 	"os"
 	"testing"
 	"time"
@@ -36,6 +37,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/client/interceptor"
 	"sigs.k8s.io/controller-runtime/pkg/envtest"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
@@ -52,6 +54,8 @@ var (
 	testEnv                 *envtest.Environment
 	k8sManager              ctrl.Manager
 	t                       *testReconciler
+	interceptorsFuncs       interceptor.Funcs
+	subResourceUpdateError  error
 
 	timeout             = "30s"
 	consistentlyTimeout = "6s"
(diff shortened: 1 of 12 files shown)


controllers/suite_test.go
```diff
@@ -81,7 +85,15 @@ var _ = BeforeEach(func() {
...
 
-	fakeK8sClient = fake.NewClientBuilder().WithScheme(scheme.Scheme).WithStatusSubresource(
+	interceptorsFuncs = interceptor.Funcs{}
+	interceptorsFuncs.SubResourceUpdate = func(ctx context.Context, client client.Client, subResourceName string, obj client.Object, opts ...client.SubResourceUpdateOption) error {
+		if subResourceUpdateError != nil {
+			return subResourceUpdateError
+		}
+		return client.Status().Update(ctx, obj, opts...)
+	}
+
```
aclevername, line 95: nice

controllers/workplacement_controller.go
```diff
@@ -107,12 +110,28 @@ func (r *WorkPlacementReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
+		workPlacement.Status.VersionID = versionID
+		err = r.Client.Status().Update(ctx, workPlacement)
+		if err != nil {
+			r.VersionCache[workPlacement.GetUniqueID()] = versionID
+			logger.Error(err, "Error updating WorkPlacement status")
+			return ctrl.Result{}, err
+		}
+	}
+	logger.Info("WorkPlacement successfully reconciled", "workPlacement", workPlacement.Name, "versionID", versionID)
+
```
aclevername, line 134: if versionID is empty and its not in the cache, this means we reconciled, set the SHA in the status, restarted the pod, and then reconciled again? Wondering if the log message is still correct in this scenario, because we cant update the status or ensure its correct

aclevername (approved):
LGTM, 1 comment but not a big issue

controllers/workplacement_controller.go
kirederik: the versionID coming from the writer is only populated once, when we push new commits/files to the repository; it will be empty whenever a reconciliation loop is executed but the workloads have not changed

if the cache is empty, it means we successfully applied to the status, so we can find the information there.

hopefully it will never be the case where:
* the resource `Status.VersionID` is empty
* the cache is empty
* the writer returned `versionID` is empty

 the only time where i think we will hit line 133 and the above will be true is when:
* new kratix is reconciling a pre-existing workplacement: we won't know the version id until we change something in the workplacement that requires a new commit

or
* we push a new commit
* we fail to set the status (which means we populate the cache)
* before the controller has a chance to reconcile the workplacement again, it dies (hence wiping the cache)
* next time the reconciliation loop runs, there's nothing in the cache and no new commits

my hope is that the latter is such an edge case that we don't need to worry about it right now

PR #156, opened 2024-06-05
Bump Go version to 1.22.4

richcooper95:
Vulnerabilities found in [this](https://app.circleci.com/pipelines/github/syntasso/kratix/1896/workflows/43358f24-e5b9-495a-bcb9-250f70be21d2/jobs/10599) CI scan, and are fixed in 1.22.4 ([release notes](https://go.dev/doc/devel/release#go1.22.minor))

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -76,7 +76,7 @@ commands:
       - k8s/install-kubectl
       - golang/install:
           cache: true
-          version: 1.22.3
+          version: 1.22.4
       - run:
           name: Install Kind
           command: |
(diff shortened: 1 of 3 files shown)


kirederik (changes requested):
changes to the go.mod don't actually bump the version of go, but rather represents the minimum version of go that's compatible with the code; and that has not changed (we are still compatible with go 1.21 iirc)

see https://go.dev/ref/mod#go-mod-file-go

to bump go, you gotta bump the version of go we use to compile kratix; so that involves changing the dockerfile and updating the circleci config.

richcooper95:
Excellent 😄 changed now, ta!

aclevername (approved):

aclevername:
dockerfile still needs to be updated @richcooper95

aclevername:
nevermind, dockerfile doesnt need bumping, i misread the version. reverting.

PR #157, opened 2024-06-07
fix: remove deprecrated 'kratix-promise-id' label

ChunyiLyu:
fix: remove deprecrated 'kratix-promise-id' label


kirederik (approved):
assuming the tests pass, all good!

PR #169, opened 2024-06-17
fix: delete kratix statefile last

ChunyiLyu:
When we cannot find .kratix state file, we can't generate the list of files to clean up for flat state store. Requeuing won't help and we need to exit

### DIFF
--- a/controllers/workplacement_controller.go
+++ b/controllers/workplacement_controller.go
@@ -51,7 +51,10 @@ type WorkPlacementReconciler struct {
 	VersionCache map[string]string
 }
 
-const repoCleanupWorkPlacementFinalizer = "finalizers.workplacement.kratix.io/repo-cleanup"
+const (
+	repoCleanupWorkPlacementFinalizer       = "finalizers.workplacement.kratix.io/repo-cleanup"
+	kratixFileCleanupWorkPlacementFinalizer = "finalizers.workplacement.kratix.io/kratix-file-cleanup"
+)
 
 //+kubebuilder:rbac:groups=platform.kratix.io,resources=workplacements,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=platform.kratix.io,resources=workplacements/status,verbs=get;update;patch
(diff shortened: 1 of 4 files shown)


controllers/workplacement_controller.go
```diff
@@ -51,7 +51,10 @@ type WorkPlacementReconciler struct {
 	VersionCache map[string]string
 }
 
-const repoCleanupWorkPlacementFinalizer = "finalizers.workplacement.kratix.io/repo-cleanup"
+const (
+	repoCleanupWorkPlacementFinalizer       = "finalizers.workplacement.kratix.io/repo-cleanup"
+	kratixFileCleanupWorkPlacementFinalizer = "finalizers.workplacement.kratix.io/kratix-file-cleanup"
```
aclevername, line 56: this is a bit generic when compared to the one above, maybe `filepath-mode-none-cleanup` or `kratix-dot-file-cleanup`, or something more descriptive?

controllers/workplacement_controller.go
```diff
@@ -129,42 +133,56 @@ func (r *WorkPlacementReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
-	_, err = writer.UpdateFiles(dir, workPlacement.Name, nil, workloadsToDelete)
 
-	if err != nil {
+	if !pendingRepoCleanup && pendingKratixFileCleanup {
+		logger.Info("cleaning up .kratix state file", "workplacement", workPlacement.Name)
+		workloadsToDelete = []string{kratixFilePath}
+		finalizerToRemove = kratixFileCleanupWorkPlacementFinalizer
+	}
+
+	if _, err := writer.UpdateFiles(dir, workPlacement.Name, nil, workloadsToDelete); err != nil {
```
aclevername, line 178: took me way to long to grok how this worked 🙈 i think codes a lot more complex just to reduce a bit of duplication. I'd rather it was just duplicated and looked ike:

```
if repoFinalierCleanuper
   ... logic
  remove finalizer
  requeue

if pendingRepoCleanup
   ... logic
   remove finalizer
   requeue
```

kirederik: i had something like this before but didn't like the duplication, hence the current state;

i could extract it to a function though... let me give it a go

aclevername (approved):

PR #179, opened 2024-06-25
refactor: move pipeline resource generation into the pipelines types

kirederik:
plus:

* remove `labels` utils; and simplified label generation
* remove left-overs/redundant labels
* simplify generation of the pipeline in the controllers
* separate the `GenerateResourceName` function to avoid import cycle

lots of changes, happy to pair-review 👀


lib/pipelineutil/job_resources.go
```diff
@@ -0,0 +1,38 @@
...
+	GetName() string
+}
+
+type pipelineObjects struct {
+	name                 string
+	job                  *batchv1.Job
+	jobRequiredResources []client.Object
+}
+
+func NewPipelineObjects(name string, job *batchv1.Job, jobRequiredResources []client.Object) PipelineJobResources {
```
ChunyiLyu, line 20: Could make this public and delete the interface.

ChunyiLyu (approved):

PR #183, opened 2024-06-27
Fix: When labels change we should update the labels on existing resources

aclevername:


controllers/promise_controller.go
```diff
@@ -498,8 +499,17 @@ func (r *PromiseReconciler) createResourcesForDynamicControllerIfTheyDontExist(c
 	err := r.Client.Create(ctx, &cr)
 	if err != nil {
 		if errors.IsAlreadyExists(err) {
-			// TODO: Handle updates of all Promise resources gracefully.
-			logger.Info("Cannot execute update on pre-existing ClusterRole")
+			existingCR := &rbacv1.ClusterRole{}
```
kirederik, line 502: should this use the controllerutil.CreateOrUpdate method instead of just create/capture error/update? 

https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#CreateOrUpdate

kirederik (approved):

PR #184, opened 2024-06-27
fix: manual-reconciliation should suspend currect job

aclevername:


kirederik (approved):
i can smell a refactor opportunity in this piece of code (three level deep ifs, with the same two conditionals repeated later....) but nothing immediate comes to mind

PR #186, opened 2024-07-03
fix: bump go

ChunyiLyu:
- there's a vulnerability in net/http


aclevername (approved):
thanks!

aclevername (changes requested):
we need to bump the go version in circleci as well

kirederik:
> we need to bump the go version in circleci as well

here-> https://github.com/syntasso/kratix/blob/3d0b967413ca6824df8939e52df5806e29254443/.circleci/config.yml#L77-L79

aclevername (approved):

PR #191, opened 2024-07-09
feat: remove versioning in OS Kratix

richcooper95:
Changes:
- Move to only have a `latest` release (and tag)
- Append the manifest digest to each of the images (Kratix and Work Creator)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -383,7 +383,7 @@ jobs:
           name: Release
           command: |
             gh config set prompt disabled
-            gh release delete <<pipeline.git.tag>>
+            gh release delete <<pipeline.git.tag>> || true
             gh release create <<pipeline.git.tag>> --title <<pipeline.git.tag>> ./distribution/**/*.yaml ./distribution/*.yaml
 
   helm-release:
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -383,13 +383,14 @@ jobs:
...
             gh config set prompt disabled
+            gh release delete <<pipeline.git.tag>>
             gh release create <<pipeline.git.tag>> --title <<pipeline.git.tag>> ./distribution/**/*.yaml ./distribution/*.yaml
 
   helm-release:
     docker:
       - image: cimg/base:current
     environment:
-      VERSION: << pipeline.git.tag >>
+      VERSION: latest
```
richcooper95, line 393: Possibly no longer need this

.circleci/config.yml
```diff
@@ -402,7 +403,7 @@ jobs:
             git add kratix/
             git diff --cached --quiet && exit 0 || true
 
-            ./scripts/bump-minor kratix/
+            VERSION=v0.0.0-latest ./scripts/bump-version kratix/
```
richcooper95, line 406: Might not need this either; we can just hard-code `v0.0.0-latest`

PR #201, opened 2024-07-19
Allow specifying custom service account

### LINKED ISSUE #196: feat: allow promise writer to set the Service Account in a pipeline
Currently, Kratix will dynamically generate the Service Account for the Pipeline following a known pattern. It's not uncommon for ours users, however, to want to control the Service Accounts in their platform themselves (see https://github.com/syntasso/kratix/issues/193).

In this story, we will introduce a new property in the Kratix pipeline to allow the Service Account name to be set. Kratix should create the SA with the provided name if it doesn't exist; if it does, it should update it with any labels or annotations we need.

Kratix deletes service accounts on certain cases, like when the promise is deleted. It should not delete the SA if it has not created it.

The Pipeline Spec should look something like:

```yaml
kind: Pipeline
metadata:
  name: pipeline-name
spec:
  rbac:
    serviceAccount: my-serviceaccount
  containers:
  - name: hello
    image: myorg/myimage
```

## Tasks
- [ ] Introduce the new field in the pipeline spec
- [ ] update docs

aclevername:
Closes #196 

- if it doesn't exist we create it
- if it does exist we use it but dont delete or modify


ChunyiLyu (approved):

lib/workflow/reconciler.go
```diff
@@ -459,6 +459,20 @@ func applyResources(opts Opts, resources ...client.Object) {
 		logger.Info("Reconciling")
 		if err := opts.client.Create(opts.ctx, resource); err != nil {
 			if errors.IsAlreadyExists(err) {
+				// if _, ok := existingResource.GetLabels()[v1alpha1.PromiseNameLabel]; !ok {
```
aclevername, line 462:
```suggestion
```

PR #204, opened 2024-07-24
Don't reconcile Works that are being deleted; set workplacement finalizers after writing files

### LINKED ISSUE #203: flakey system test
https://app.circleci.com/pipelines/github/syntasso/kratix/2122/workflows/d3be1b25-8cbb-48e4-b51b-4ae8083bde3f/jobs/11711

https://app.circleci.com/pipelines/github/syntasso/kratix/2085/workflows/e930bcae-1e59-43b7-8b40-c50516eecc65/jobs/11551

failing every 5/6 runs

aclevername:
closes #203 

green runs (the red is a different testing flaking, I'm investigating and will address in a different PR) 
Screenshot 2024-07-24 at 14 11 39

this bug was pretty nasty to debug, and highlighted two issues:
- Destination controller triggers the reconciliation of ALL works, even if the work in question is being deleted, which causees it to recreate a workplacement immediately after its been deleted (race conditon)
- A workplacement gets the finalizer added for deleting the files it creates *before* it creates them, which means if a workplacement is created, finalizers added (requeue triggered), workplacement deleted, requeue starts the deletion path despite no files having been created.

## Example failure

Chronological order:

1. An existing healthy work get deleted, triggers deletion of workplacement
1  Work being deleted

2. Workplacement gets successfully deleted
2  successfull deletion of wp

3. The destination controller then triggers a reconcilation of all works (via the code directly, not via the work controller), the Work placement which is being deleted hasn't been deleted *yet*, because there is a time delay between the workplacement successfully being deleted, and the work resource having its finalzier removed. This reconciliation therefore then creates a new workplacement for the work
3  destination triggering reconciliation of deleting work resource

4. Workplacement gets created with finalizers, and then requeues (no files written yet!)
4  workplacement recreated, adding finalizers
 Work being deleted](https://github.com/user-attachments/assets/bcad7c45-086e-4c10-82db-bb120a8a552f)
 
5. The workplacement gets marked as deleted, even though its not created any files yet
5  deletion triggered- before workplacement actualy writes files

6. Workplacement deletion fails because the finalizers exist, but the files do not
6  workplacement deletion fails because it never wrote the fails, but has the finalizers

### DIFF
--- a/controllers/work_controller.go
+++ b/controllers/work_controller.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+
 	"github.com/go-logr/logr"
 	"github.com/syntasso/kratix/api/v1alpha1"
 	"k8s.io/apimachinery/pkg/api/errors"
@@ -25,6 +26,8 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
 const workCleanUpFinalizer = v1alpha1.KratixPrefix + "work-cleanup"
@@ -124,5 +127,30 @@ func (r *WorkReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&v1alpha1.Work{}).
 		Owns(&v1alpha1.WorkPlacement{}).
+		Watches(
+			&v1alpha1.Destination{},
+			handler.EnqueueRequestsFromMapFunc(r.requestReconilationOfAllWorksOnDestinationCreateOrUpdate),
+		).
 		Complete(r)
+}
+
+func (r *WorkReconciler) requestReconilationOfAllWorksOnDestinationCreateOrUpdate(ctx context.Context, obj client.Object) []reconcile.Request {
+	dest := obj.(*v1alpha1.Destination)
+
+	if dest.GetDeletionTimestamp() != nil {
+		return nil
+	}
+
+	allWorks := &v1alpha1.WorkList{}
+	err := r.Client.List(ctx, allWorks)
+	if err != nil {
+		r.Log.Error(err, "Error listing all Works")
+		return nil
+	}
+
+	var requests []reconcile.Request
+	for _, work := range allWorks.Items {
+		requests = append(requests, reconcile.Request{NamespacedName: client.ObjectKeyFromObject(&work)})
+	}
+	return requests
 }
--- a/controllers/workplacement_controller.go
+++ b/controllers/workplacement_controller.go
@@ -73,6 +73,8 @@ func (r *WorkPlacementReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		return defaultRequeue, nil
 	}
 
+	logger.Info("Reconciling WorkPlacement")
+
 	destination := &v1alpha1.Destination{}
 	destinationName := client.ObjectKey{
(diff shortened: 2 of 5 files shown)


controllers/workplacement_controller.go
```diff
@@ -120,15 +119,20 @@ func (r *WorkPlacementReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
 			return ctrl.Result{}, err
 		}
 	}
-	logger.Info("WorkPlacement successfully reconciled", "workPlacement", workPlacement.Name, "versionID", versionID)
 
+	if missingFinalizers := checkWorkPlacementFinalizers(workPlacement, filepathMode); len(missingFinalizers) > 0 {
+		return addFinalizers(opts, workPlacement, missingFinalizers)
+	}
+
+	logger.Info("WorkPlacement successfully reconciled", "workPlacement", workPlacement.Name, "versionID", versionID)
```
aclevername, line 135: the order is now:

1. update the files in statestore
1. update the status of the workplacement with the latest VersionID
1. update the finalizers

I considered swapping 2/3 round, i dont mind tbh. They both can't be done at the same time annoyingly since its an update to the status and the metadata, which is two separate API calls, which won't work without requeuing since the resource will have changed between the two API calls

lib/writers/git.go
```diff
@@ -219,7 +219,7 @@ func (g *GitWriter) ReadFile(filePath string) ([]byte, error) {
 	defer os.RemoveAll(filepath.Dir(localTmpDir))
 
 	if _, err := worktree.Filesystem.Lstat(fullPath); err != nil {
-		logger.Error(err, "could not stat file")
+		logger.Info("could not stat file", "err", err)
```
aclevername, line 222: sometimes this is fine, so having this error output was making me think the reconciler exited when it hasn't. the cosnume of this detects if it should error log or ignore

controllers/workplacement_controller_test.go
```diff
@@ -165,7 +165,7 @@ var _ = Describe("WorkplacementReconciler", func() {
 				Expect(result).To(Equal(ctrl.Result{}))
 
 				By("calling UpdateFiles()")
-				Expect(fakeWriter.UpdateFilesCallCount()).To(Equal(1))
+				Expect(fakeWriter.UpdateFilesCallCount()).To(Equal(2))
```
ChunyiLyu, line 168: Why is this needed?

controllers/work_controller.go
```diff
@@ -124,5 +127,30 @@ func (r *WorkReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+
+func (r *WorkReconciler) requestReconilationOfAllWorksOnDestinationCreateOrUpdate(ctx context.Context, obj client.Object) []reconcile.Request {
+	dest := obj.(*v1alpha1.Destination)
+
+	if dest.GetDeletionTimestamp() != nil {
+		return nil
+	}
+
+	allWorks := &v1alpha1.WorkList{}
+	err := r.Client.List(ctx, allWorks)
```
ChunyiLyu, line 145: `scheduler.ReconcileAllDependencyWorks()` only reconciles if `work.IsDependency()` is true. Should we have the same check in here?

controllers/workplacement_controller_test.go
aclevername: the reconciliation loop runs twice, because the finalizer is now after the update an additional call is made to the update. Before it was finazlers->requque->finalizers->update->done, now its update->requeue->update->finalizers->done

controllers/work_controller.go
aclevername: I don't think so. If a work is for a resource request and is not scheduled (no matching destination) we do want to reconcile the work. Also if a Work for a resource request is scheduled to a destination, and then that destination is updated to no longer match the labels that made it get scheduled to by the work to begin with we update the work status.

ChunyiLyu (approved):

PR #208, opened 2024-07-24
Don't reconcile all RRs on promise creation; enforce a 1 second period between job creations

aclevername:
new green runs:
Screenshot 2024-07-25 at 12 11 44

We had a series of race conditions causing some test failures:
https://app.circleci.com/pipelines/github/syntasso/kratix/2135/workflows/db157571-5e8d-4062-9d77-cb0a95fbea28/jobs/11979
https://app.circleci.com/pipelines/github/syntasso/kratix/2133/workflows/a53d9571-223d-477e-a791-bb9e673b864c/jobs/11884

the observed behaviour of these failures were always: 
- 2 jobs for the 1st pipeline (1 suspended so no pod created)
-  N (N>1) jobs for the 2nd pipeline
- all these pods/jobs having the same creation timestamp (within 1 second of eachother)
Screenshot 2024-07-24 at 16 04 10

Three issues:
- When a Promise Generation changes we trigger the manual reconciliation of all RRs, this was happening on the Creation event (generation=1), which meant it was suspending jobs unnecessarily. If you very quickly created a Promise, then RR, we could accidentally create two jobs (suspending one of them)
- We handled the scenario of a Job being suspended as it being finished successfully, which was incorrect
- We can create multiple jobs within a second of each other, meaning they have the same creationTimestamp, which causes issues in our code as we rely on sorting jobs by time order to determine the latest one. This also led to race conditions as our k8s client uses a cache, meaning that if the reconciler ran repeatedly within a few milliseconds of eachother, we would create a load of jobs because the cache didn't list any jobs as existing, because it doesn't always give the most up to date information.

fixes:
- Gracefully handle the `promise.metadata.generation`==1 scenario
- Introduce a 1 second delay after creating jobs. Each controller runs reconciliations sequentially, and using the requeue isn't sufficient since external things can trigger the reconciler to run again (ignoring the desired requeue time)
- Handle suspended jobs as a failure


lib/workflow/export_test.go
```diff
@@ -0,0 +1,7 @@
+package workflow
```
ChunyiLyu, line 1:
```suggestion
package workflow_test
```
Shouldn't this helper be part of the test package?

lib/workflow/export_test.go
```diff
@@ -0,0 +1,7 @@
...
+
+import "time"
+
+func SetMinimumPeriodBetweenCratingPipelineResources(t time.Duration) {
```
ChunyiLyu, line 5:
```suggestion
func SetMinimumPeriodBetweenCreatingPipelineResources(t time.Duration) {
```

typo

lib/workflow/reconciler.go
```diff
@@ -29,6 +32,8 @@ type Opts struct {
 	source    string
 }
 
+var minimumPeriodBetweenCreatingPipelineResources = 1100 * time.Millisecond
```
ChunyiLyu, line 35:
```suggestion
var minimumPeriodBetweenCreatingPipelineResources = 1 * time.Second
```
why not 🤔

aclevername: because I'm paranoid

aclevername: it needs to be *atleast* 1 second, so I went slightly above 🤷

ChunyiLyu (approved):

Issue #227, opened 2024-08-27
Feat: Provide default security context for all Kratix related pods to use


abangser:
Users require the ability to set a `securityContext` for any workload running in their cluster which means some Kratix managed jobs/pods/containers must adhere to this context.

At a minimum, we need a way for users to declare their default `securityContext`. We may choose to provide overrides for certain workflow containers but that can be put into a separate piece of work if that makes more sense to do.

When working on this, be aware that Kratix runs a number of workloads/containers including (but maybe not limited to):
* Controller manager
* Reader in workflows
* Status writer in workflows
* Work writer in workflows
* Workflow Jobs -> Pods

These should all be solved by setting [`securityContext`](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) on the pods we own. We may then choose to provide `securityContext` on the pipeline containers so that people can override as they want.

abangser:
Baseline security context per https://kubernetes.io/docs/concepts/security/pod-security-standards/#baseline:

```
                securityContext:
                    capabilities:
                      drop:
                      - ALL
                    privileged: false
                    runAsNonRoot: true
                    seccompProfile:
                      type: RuntimeDefault
```

aclevername:
Work left:
- Use the defaultSecurityContext for the user provided pods
- Allow the setting of the value for the defaultSecruityContext
  - If the user sets one in the Promise, this should take priority over the globally set defaultSecurityContext

shano:
- Fixed issue with defaultSecuriy contrxts
- Need to write tests: 1. to test the logic and 2. to ensure the overrides are working.

aclevername:
Closed by #229

PR #229, opened 2024-08-29
feat: Run Kratix with default securityContext and allow users to set their own default at Promise or Kratix level

abangser:
Related issue #227

This PR resolves 3 issues:
- Sets a sensible default for all Kratix containers (reader/work-writer/status-updater) in the pipeline
- Lets the user configure a security context for a given container the pipeline spec in a Promise
- Lets the user configure a global default security context for the users containers via a global  Kratix configmap
  - This configmap is optional, and is therefore loaded at runtime instead of as a volume mount.
  
  docs: https://github.com/syntasso/kratix-docs/pull/109


api/v1alpha1/pipeline_types.go
```diff
@@ -53,6 +54,19 @@ const (
...
+		Privileged:   ptr.To(false),
+		Capabilities: &corev1.Capabilities{
+			Drop: []corev1.Capability{"ALL"},
+		},
+		SeccompProfile: &corev1.SeccompProfile{
+			Type: "RuntimeDefault",
+		},
+	}
+)
+
```
abangser, line 69: @catmo-syntasso  Something to think about here:

Do we want the users to be able to set a securityContext for Kratix owned resources (e.g. the containers we add to the pipelines)? In one way, that would make sense as they may have requirements to run containers in their clusters. In another way, that would mean Kratix would be subject to any rules they apply which may or may not be possible with the current code.

@shano and I were discussing and would recommend we stick with a Kratix default (we have picked a sensible one that is default for Kyverno, suggested by Kubernetes docs, etc). This will let us move faster with Kratix without big risk as changing to user provided would be a very small change.

main.go
```diff
@@ -226,3 +230,16 @@ func getNumJobsToKeep() int {
 	}
 	return numberOfJobsToKeep
 }
+
+func getDefaultSecurityContext() *corev1.SecurityContext {
+	return &corev1.SecurityContext{
```
abangser, line 235: Note: this is just to stub out the behaviour as we codify using the context. We will need to capture the user input here or set to empty context.

PR #237, opened 2024-09-10
refactor: move jobs cleanup limit to kratix config

ChunyiLyu:
## Context
Number of jobs to keep used to be configured via env var. This PR moves it to kratix configmap. It can be configured like so:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: kratix
data:
  config: |
    numberOfJobsToKeep: 3
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -21,7 +21,6 @@ import (
 	"flag"
 	"fmt"
 	"os"
-	"strconv"
 	"time"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
@@ -59,7 +58,8 @@ func init() {
 }
 
 type KratixConfig struct {
-	Workflows Workflows `json:"workflows"`
+	Workflows          Workflows `json:"workflows"`
+	NumberOfJobsToKeep int       `json:"numberOfJobsToKeep,omitempty"`
 }
 
 type Workflows struct {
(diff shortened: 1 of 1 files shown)


aclevername:
@abangser configmaps only allow top-level keys, e.g.

```
data:
  foo: bar
```

You can't do this:
```
data:
  foo:
    bar:
      baz: ban
```

@ChunyiLyu also pointed out its only strings as well, no other types. I've seen the `config` key equal to yaml/json/whatever pattern a lot before in the k8s world (but maybe a nicer new way has appeared)

main.go
```diff
@@ -268,18 +268,16 @@ func readKratixConfig(logger logr.Logger, kClient client.Client) (*KratixConfig,
...
+		if kratixConfig.NumberOfJobsToKeep < 1 {
+			setupLog.Error(fmt.Errorf("invalid Kratix Config"),
+				"numberOfJobsToKeep cannot be less than one; set to default value",
+				"numberOfJobsToKeep", kratixConfig.NumberOfJobsToKeep)
 			numberOfJobsToKeep = numJobsToKeepDefault
+		} else {
+			numberOfJobsToKeep = kratixConfig.NumberOfJobsToKeep
 		}
 	}
 	return numberOfJobsToKeep
```
kirederik, line 283: In my crusade against `else`, wdyt about:

```go
if kratixConfig == nil || kratixConfig.NumberOfJobsToKeep == 0 {
    return numJobsToKeepDefault
}

if kratixConfig.NumberOfJobsToKeep < 1 {
    setupLog.Error(fmt.Errorf("invalid Kratix Config"),
        "numberOfJobsToKeep cannot be less than one; set to default value",
        "numberOfJobsToKeep", kratixConfig.NumberOfJobsToKeep)
    return numJobsToKeepDefault
}

return kratixConfig.NumberOfJobsToKeep
```

👀

ChunyiLyu:
@kirederik @abangser Would you mind taking a look again and let me know is there anything missing?

kirederik (approved):

PR #241, opened 2024-09-23
feat: propagate workflow labels to pod

kirederik:
- **feat: propagate pipeline labels/annot to pod**
- **feat: add webhook for workflow labels validation**


abangser (approved):
Very nice. And appreciate the webhook for faster feedback 🙌 

I have reviewed based on screenshots/code review so worth checking this e2e with the azure use case that prompted it. Will just need to make sure to use the annotated service account as well.

PR #251, opened 2024-10-04
feat: configurable re-reconciliation of Kratix resources

richcooper95:


config/rbac/role.yaml
```diff
@@ -28,7 +28,7 @@ rules:
   verbs:
   - get
   - list
-  - watch
+  - watch // Also used by promiseRelease
```
abangser, line 31: I _think_ this is what caused issues on another commit when it got to the helm chart. Let's review before merging.

PR #261, opened 2024-10-28
chore: argo applications setup

ChunyiLyu:
## Context

Example on argo applications that reconciles on `resources/` and `dependencies/`.


abangser (approved):
LGTM - This matches how we set up the [applications in the Konstruct gitops catalog](https://konstruct.io/kubefirst) (for use with [kubefirst](https://konstruct.io/kubefirst)).

Issue #266, opened 2024-10-31
bug: rbac resourceNamespace does not work with resource requests created in different namespaces


ChunyiLyu:
## Description

When:
1. using pipeline permission in resource workflow pipelines
2. setting `resourceNamespace`, whether to '*' or a specific namespace
3. create multiple resource requests from different namespace

There should be one rolebinding or one clusterrolebinding created for *each* namespace that resource request is created in. However, kratix only ever creates one binding.

## To reproduce

1. install a promise with pipeline permission set
```yaml

apiVersion: platform.kratix.io/v1alpha1
kind: Promise
metadata:
  creationTimestamp: null
  name: rbac
spec:
  api:
    apiVersion: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    metadata:
      creationTimestamp: null
      name: tests.syntasso.io
    spec:
      group: syntasso.io
      names:
        kind: test
        plural: tests
        singular: test
      scope: Namespaced
      versions:
        - name: v1alpha1
          schema:
            openAPIV3Schema:
              properties:
                spec:
                  properties:
                    test:
                      type: string
                  type: object
              type: object
          served: true
          storage: true
  workflows:
    promise: {}
    resource:
      configure:
        - apiVersion: platform.kratix.io/v1alpha1
          kind: Pipeline
          metadata:
            name: create
          spec:
            rbac:
              permissions:
                - apiGroups: [""]
                  verbs: ["list"]
                  resources: ["deployment"]
                  resourceNamespace: "*"
            containers:
              - image: ghcr.io/syntasso/kratix-pipeline-utility:v0.0.1
                name: test
                ...
```
2. create resource in different namespaces
```yaml
apiVersion: syntasso.io/v1alpha1
kind: test
metadata:
  name: zero
  namespace: default
---
apiVersion: syntasso.io/v1alpha1
kind: test
metadata:
  name: one
  namespace: test
```
3. list created clusterrolebinding
```
 k get clusterrolebinding | grep create
rbac-resource-configure-create-default-992d2           ClusterRole/rbac-resource-configure-create-kratix-all-namespaces-8dd84
```
4. see that there is only one clusterrolebinding created for one of the namespace. 
Expectation is that there should be two: one to binding to service account in namespace 'test', the other one bind to the service account in 'default' namespace.

## Issue

clusterrolebinding and rolebinding name is not pipeline namespace.
[...]

PR #268, opened 2024-11-04
fix: create separate bindings for resources requests from different namespaces with user provided RBAC

aclevername:
Closes: #266 

This PR:
- generates role binding and cluster role binding names for user provided permission based on pipeline(or service account) namespace rather than the namespace that the rbac is for
- add a new resource namespace label to user provided permission object to distinguish bindings created for service accounts in different namespaces when deleting
- created a `pipeline_factory.go` for `PipelineFactory` and its methods
- adds a system test creating one resource request from default namespace, one from 'test' namespace.

### DIFF
--- a/test/system/system_test.go
+++ b/test/system/system_test.go
@@ -285,13 +285,13 @@ var _ = Describe("Kratix", func() {
 				platform.eventuallyKubectl("get", "crd", crd.Name)
 				worker.eventuallyKubectl("get", "namespace", declarativeWorkerNamespace)
 
-				platform.kubectlForce("delete", "namespace", "pipeline-perms-ns")
+				platform.eventuallyKubectlDelete("namespace", "pipeline-perms-ns")
 				platform.eventuallyKubectl("create", "namespace", "pipeline-perms-ns")
 			})
 
 			It("executes the pipelines and schedules the work to the appropriate destinations", func() {
 				rrName := bashPromiseName + "rr-test"
-				platform.kubectl("apply", "-f", exampleBashRequest(rrName, "old"))
+				platform.kubectl("apply", "-f", exampleBashRequest(rrName, "default", "old"))
 
 				oldRRDeclarativePlatformNamespace := "declarative-platform-only-" + rrName + "-old"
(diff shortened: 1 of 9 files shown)


test/system/system_test.go
```diff
@@ -548,6 +547,58 @@ var _ = Describe("Kratix", func() {
...
+				platform.eventuallyKubectl("get", "crd", crd.Name)
+				worker.eventuallyKubectl("get", "namespace", declarativeWorkerNamespace)
+				platform.kubectl("create", "ns", "test")
+			})
+
+			AfterEach(func() {
+				platform.kubectl("delete", "ns", "test")
+			})
+
+			It("works", func() {
```
aclevername, line 563:
```suggestion
			It("creates uniques RBAC resources for both requests, including cross-namespaces resources", func() {
```

something like that?

test/system/system_test.go
```diff
@@ -133,13 +132,13 @@ var _ = Describe("Kratix", func() {
...
-			platform.eventuallyKubectlDelete("clusterrole", bashPromiseName+"-default-resource-pipeline-credentials")
-			platform.eventuallyKubectlDelete("clusterrolebinding", bashPromiseName+"-default-resource-pipeline-credentials")
-			platform.eventuallyKubectlDelete("serviceaccount", bashPromiseName+"-existing-custom-sa")
+			platform.eventuallyKubectlDelete("promisereleases", bashPromiseName, "default")
+			platform.eventuallyKubectlDelete("promise", bashPromiseName, "default")
+			platform.eventuallyKubectlDelete("namespace", imperativePlatformNamespace, "default")
+			platform.eventuallyKubectlDelete("namespace", "imperative-"+bashPromiseName+"-test", "default")
+			platform.eventuallyKubectlDelete("clusterrole", bashPromiseName+"-default-resource-pipeline-credentials", "default")
+			platform.eventuallyKubectlDelete("clusterrolebinding", bashPromiseName+"-default-resource-pipeline-credentials", "default")
+			platform.eventuallyKubectlDelete("serviceaccount", bashPromiseName+"-existing-custom-sa", "default")
```
aclevername, line 141: why do we need this?

PR #298, opened 2024-12-11
feat: don't reconcile resource requests while a Promise pipeline is in-flight

### LINKED ISSUE #279: feat: Don't reconcile resource requests while a Promise pipeline is in-flight
Related to #269, when a Promise is updated or the Promise pipeline is manually triggered, things in the system may change as a result of this. In #269 , it was the destination selectors. If we allow resource requests to be reconciled while this change occurring we might get odd behaviour, such as:
```yaml
Promise:
  spec:
    destinationSelectors:
    - env: dev
   workflows:
     resource.configure.image: v0.1.0
```
Updated to:
```
Promise:
  spec:
    destinationSelectors:
    - env: **prod**
   workflows:
     resource.configure.image: **v0.2.0**
```

We could potentially end up running a resource request pipelines with image `v0.2.0` and destination selectors `env: dev`, which is bad. Eventually the resource request would rerun with `v0.2.0` and `env: prod`, but being this in-between state is potentially dangerous.

alparslanavci:
Closes #279.

### DIFF
--- a/controllers/promise_controller.go
+++ b/controllers/promise_controller.go
@@ -239,6 +239,12 @@ func (r *PromiseReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 
 	if ctrlResult != nil {
 		logger.Info("stopping reconciliation while reconciling dependencies")
+		if r.dynamicControllerHasAlreadyStarted(promise) {
+			logger.Info("dynamic controller already started, marking it not to create resources")
+			dynamicController := r.StartedDynamicControllers[string(promise.GetUID())]
+			dynamicControllerCanCreateResources := false
+			dynamicController.CanCreateResources = &dynamicControllerCanCreateResources
+		}
 		return *ctrlResult, nil
 	}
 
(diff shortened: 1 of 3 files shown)


kirederik:
hey @alparslanavci, thanks for the PR. The code looks good and, apart from what you raised in #301, nothing much to worry here. We could try to avoid the potential race condition by making the dynamic controller verify the promise conditions (it has to fetch the promise anyway)-- so that might be a better approach.

however, what we are actually debating is whether we want this feature at all. Gotta chat with the team and get back to you here!

Thanks again!

alparslanavci:
Thanks for the review, @kirederik! I agree with verifying the promise conditions on the dynamic resource controller. It would be a better approach. 

No worries about the feature discussion, just let me know if you want to close the PR. 👍

alparslanavci:
BTW, it might be better to add a new flag on `Promise` to leave the "reconcile resource requests while a Promise pipeline is in-flight" decision to the user. This will allow the user to opt between availability and consistency.

kirederik:
Alright!

We debated a bit and decided that this is not really something we should incorporate into the product right now.

I'm going to close this PR for the moment. We will get back to it if things change!

Thanks again

PR #300, opened 2024-12-16
Schedule a healthDefinition when resource healthchecks is set

ChunyiLyu:
## Context

closes https://github.com/syntasso/kratix-private/issues/97

#### Updates
1. added `spec.healthchecks` to Promise schema. Example resource healthcheck in promise:
```yaml
kind: Promise
metadata:
  name: redis
spec:
  healthChecks:
    resource:
      schedule: 2h
      workflow:
        kind: Pipeline
        metadata:
          name: health
          namespace: default
        spec:
          containers:
            - image: ghcr.io/syntasso/kratix-marketplace/redis-configure-pipeline:v0.1.0
              name: redis-healthcheck

  api: ...
  workflows: {}
```
2. when `spec.healthchecks.resource` is set, an addition job will be created when resource is requested, which will output a single `healthdefinition.yaml`
`healthdefinition.yaml` is created with user input and by pipeline adaptor image. This is done by `health-definition-creator` script: https://github.com/syntasso/kratix/blob/healthdefinition/work-creator/scripts/health-definition-creator. It's currently a bash script because `HealthDefinition` type hasn't been defined yet. Implementing a go binary likely means yaml templating. There's a follow up chore to refactor to a go binary once `HealthDefinition` has been defined
3. Added a system test to make sure we are running an additional job and schedule `healthdefinition.yaml`. Test is intentionally minimal in assertion and skipped by default. Once `HealthDefinition` is defined, we could assert on the end-to-end flow which is more aligned with the nature of our system test.

Follow up chore on the bash script:
https://github.com/syntasso/kratix-private/issues/105


api/v1alpha1/pipeline_factory.go
```diff
@@ -89,7 +89,8 @@ func (p *PipelineFactory) serviceAccount() *corev1.ServiceAccount {
 }
 
 func (p *PipelineFactory) configMap(workloadGroupScheduling []WorkloadGroupScheduling) (*corev1.ConfigMap, error) {
-	if p.WorkflowAction != WorkflowActionConfigure {
+	// used to be p.WorkflowAction != WorkflowActionConfigure
```
kirederik, line 92: we can prob get rid of this comment (and the one below)

PR #305, opened 2024-12-18
chore: use the term 'pipeline-adapter' instead of 'wc-img'

SaphMB:


kirederik:
Changes look good, thanks!

We need to reproduce the same change on the ske-operator, so it can deal with the new env var name as well

PR #319, opened 2025-01-09
chore: rename all references to kratix-bucket to kratix

abangser:
this is a cross repo change and changes in the docs, marketplace, and enterprise repos depend on this.


abangser:
Confirmed by @SaphMB 🙌

PR #322, opened 2025-01-13
chore: Add support for argocd in install-gitops script

aclevername:


scripts/install-gitops
```diff
@@ -77,9 +81,7 @@ patch_flux_resources() {
         sed "s/kratix-workload-resources/kratix-${name}-resources/g"
 }
 
-install_gitops() {
-    CONTEXT="${CONTEXT:-$1}"
-    BUCKET_PATH="${BUCKET_PATH:-$2}"
+install_flux_gitops() {
```
SaphMB, line 84:
```suggestion
install_flux_gitops() {
    CONTEXT="${CONTEXT:-$1}"
    BUCKET_PATH="${BUCKET_PATH:-$2}"
```

To ensure that the behaviour in quick-start is consistent, at present it will install the flux resources on the `kind-platform`

scripts/install-gitops
```diff
@@ -9,6 +9,7 @@ cd $ROOT
 CONTEXT=""
 WORKER_STATESTORE_TYPE=${WORKER_STATESTORE_TYPE:-BucketStateStore}
 KUSTOMIZATION_NAME=""
+GITOPS_PROVIDER="flux"
```
SaphMB, line 12:
```suggestion
GITOPS_PROVIDER=${GITOPS_PROVIDER:-"flux"}
```

To ensure the GITOPS_PROVIDER can be overridden when calling the script

SaphMB (changes requested):
Just two little notes, suggestions in comments

- Attempted to validate the default behaviour or quick start by running `make quick-start` but reconciliation of the namespace was unsuccessful as the flux resources were being created on the wrong cluster
- Attempted to validate the use of the argo path by running `GITOPS_PROVIDER=argo ./scripts/install-gitops --context kind-worker --git` but observed that the env var was being reset to flux

scripts/install-gitops
aclevername: dam i fixed this locally and forgot to push 🙈 sorry!

scripts/quick-start.sh
```diff
@@ -238,15 +238,15 @@ setup_worker_destination() {
         ${ROOT}/scripts/register-destination --name platform-cluster --context kind-platform $flags
     else
         cat "${ROOT}/config/samples/platform_v1alpha1_worker.yaml" | patch_statestore | kubectl --context kind-platform apply --filename -
-        install_gitops kind-worker worker-1
+        install_flux_gitops kind-worker worker-1
```
abangser, line 241: Not sure you want to block this PR, but should we have an Argo flag for the worker gitops too?

SaphMB (approved):

scripts/quick-start.sh
aclevername: eventually yes!

PR #337, opened 2025-01-28
Introduce linting with golangci

shano:
## Summary

This PR adds Golang linting config and Make targets.

The new targets are:

- `make lint-required` -- uses `.golangci-required.yml`, a mandatory and currently-green set of config which will run in CI
- `make lint-changed` -- uses the default `.golangci.yml`, a more thorough set of config, which runs on changed files locally (both staged and unstaged)

### `.golangci.yml`

The default config which the `golangci-lint` CLI uses, and which is also used by IDE plugins by default.

Pretty much taken verbatim from [this GitHub gist](https://gist.github.com/maratori/47a4d00457a92aa426dbd48a18776322). This is quite opinionated and is intended to be aspirational.

It's very open to changes, and is currently red 🔴

### `.golangci-required.yml`

A required subset of config which will run in CI, intended to include config which:
1. catches issues which are useful, important, and/or trivial-but-nice for future work
2. is either passing already, or easy to make pass by fixing small issues (to reduce the scope of this initial work)

Starting from this baseline green config, we can iteratively improve the codebase (e.g. adding more linters, reducing the limits for cyclomatic/cognitive complexity, etc.).

This config has already caught some bugs!

Open to changes, and is currently green 🟢

> [!NOTE]  
> Linter config is difficult to balance when introducing it fresh into a codebase; everyone has different opinions on what's important/useful/trivial. With this in mind, we're expecting these config files to change quite a lot at the start, and evolve over time. This is just the starting point! 👀

## Commits

The first commit introduces the Make targets, and the baseline config files. Each subsequent commit adds a new linter to `.golangci-required.yml` and fixes any issues which were caught by it.

Reviewing commit-by-commit is recommended!

### DIFF
--- a/controllers/destination_controller.go
+++ b/controllers/destination_controller.go
@@ -179,11 +179,11 @@ func (r *DestinationReconciler) createDependenciesPathWithExample(writer writers
 func (r *DestinationReconciler) deleteDestination(o opts, destination *v1alpha1.Destination, writer writers.StateStoreWriter) (ctrl.Result, error) {
 	if controllerutil.ContainsFinalizer(destination, destinationCleanupFinalizer) {
 		if success, err := r.deleteDestinationWorkplacements(o, destination); !success || err != nil {
-			return defaultRequeue, nil
+			return defaultRequeue, nil //nolint:nilerr // requeue rather than exponential backoff
 		}
 
 		if err := r.deleteStateStoreContents(o, writer); err != nil {
-			return defaultRequeue, nil
+			return defaultRequeue, nil //nolint:nilerr // requeue rather than exponential backoff
 		}
 
 		controllerutil.RemoveFinalizer(destination, destinationCleanupFinalizer)
--- a/controllers/destination_controller.go
+++ b/controllers/destination_controller.go
@@ -18,8 +18,9 @@ package controllers
 
 import (
 	"context"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 	"path/filepath"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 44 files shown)


controllers/promise_controller.go
```diff
@@ -132,7 +132,7 @@ func (r *PromiseReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	}
 	if client.IgnoreNotFound(err) != nil {
 		r.Log.Error(err, "Failed getting Promise", "namespacedName", req.NamespacedName)
-		return defaultRequeue, nil
+		return defaultRequeue, nil //nolint:nilerr // TODO: Do we want to return the error here?
```
richcooper95, line 135: This caused tests to fail, so need to check the right behaviour

controllers/promise_controller.go
```diff
@@ -744,7 +743,7 @@ func (r *PromiseReconciler) deletePromise(o opts, promise *v1alpha1.Promise) (ct
 		o.logger.Info("deleting resources associated with finalizer", "finalizer", dynamicControllerDependantResourcesCleanupFinalizer)
 		err := r.deleteDynamicControllerAndWorkflowResources(o, promise)
 		if err != nil {
-			return defaultRequeue, nil
+			return defaultRequeue, nil //nolint:nilerr // TODO: Do we want to return the error here?
```
richcooper95, line 746: This caused tests to fail, so need to check the right behaviour

controllers/destination_controller.go
```diff
@@ -178,11 +179,11 @@ func (r *DestinationReconciler) createDependenciesPathWithExample(writer writers
 func (r *DestinationReconciler) deleteDestination(o opts, destination *v1alpha1.Destination, writer writers.StateStoreWriter) (ctrl.Result, error) {
 	if controllerutil.ContainsFinalizer(destination, destinationCleanupFinalizer) {
 		if success, err := r.deleteDestinationWorkplacements(o, destination); !success || err != nil {
-			return defaultRequeue, nil
+			return defaultRequeue, err
```
richcooper95, line 182: This didn't cause tests to fail, but may still not be correct

controllers/destination_controller.go
```diff
@@ -178,11 +179,11 @@ func (r *DestinationReconciler) createDependenciesPathWithExample(writer writers
...
 		}
 
 		if err := r.deleteStateStoreContents(o, writer); err != nil {
-			return defaultRequeue, nil
+			return defaultRequeue, err
```
richcooper95, line 186: This didn't cause tests to fail, but may still not be correct

controllers/destination_controller.go
shano: Is it worth no-linting with a comment for now?

controllers/promise_controller.go
richcooper95: See https://github.com/syntasso/kratix/pull/337#discussion_r1935389134

PR #343, opened 2025-01-31
feat/330/add destination status

### LINKED ISSUE #330: feat: Set status and publish events in Destination to indicate if test files are successfully written
## Context

Destination should have status field to indicate to users feedback whether the creation is successful. 

We write canary resource files, a namespace and a configmap, to Destinations. Provide user feedback in the status on whether the write is successful.

```gherkin
Given user has created a Destination
Given this Destination references an existing StateStore that can be written to
Then Destination has a Status field telling users that writing canary files were successful
Then Kratix publishes a successful write event to the Destination

Given user has created a Destination
When this Destination references a StateStore that kratix cannot write to
Then users are informed in kratix logs, Destination Status, and events of the Destination
```

kirederik:
- **feat(#330): add destination status and events**
- **chore: improve test readability and coverage**
- **feat: destination to reconcile on store updates**
- **chore(#330): add system-tests for destination**

this PR closes #330


api/v1alpha1/destination_types.go
```diff
@@ -105,13 +105,13 @@ func (d *Destination) GetCleanup() string {
...
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
 }
 
-//+kubebuilder:object:root=true
-//+kubebuilder:subresource:status
-//+kubebuilder:resource:scope=Cluster,path=destinations,categories=kratix
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Cluster,path=destinations,categories=kratix
+// +kubebuilder:printcolumn:name="Ready",type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].status`,description="Indicates the destination is ready to use"
```
ChunyiLyu, line 115: nice touch for the additional printed column!

ChunyiLyu (approved):
LGTM! 

The only additional thing is that do you think we could check the existence of statestore in the destination webhook and fail early? Is that beneficial for user experience or does that creates a hard ordering dependency of destination -> statestore.

kirederik:
> The only additional thing is that do you think we could check the existence of statestore in the destination webhook and fail early? Is that beneficial for user experience or does that creates a hard ordering dependency of destination -> statestore.

Good shout... I think i'm more inclined to not force the ordering though; that's would also be consistent with other k8s resources? like if you have a deployment mounting a configmap that does not exist it accepts the creation just fine.

That said, a `Warning: state store referenced does not exist` wouldn't hurt. Maybe as a separate story?

ChunyiLyu:
@kirederik a warning sounds like a nice middle ground. Issue: https://github.com/syntasso/kratix/issues/345

PR #344, opened 2025-01-31
chore: simply comparison logic

ChunyiLyu:
## Context

Use DeepEqual instead of comparing field by field. 

checking `len(oldReqs)+len(newReqs) == 0` because a non-nil empty slice and a nil slice (for example, []byte{} and []byte(nil)) are not deeply equal, https://pkg.go.dev/reflect#DeepEqual. In this specific case `promise.Status.RequiredPromises` when not set is `nil`, but the computed new required promises is initialized to `[]v1alpha1.RequiredPromiseStatus{}` and therefore won't be equal.

### DIFF
--- a/controllers/promise_controller.go
+++ b/controllers/promise_controller.go
@@ -19,14 +19,13 @@ package controllers
 import (
 	"context"
 	"fmt"
-	"slices"
+	"reflect"
 	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


kirederik:
LGTM

although the function should probably be renamed... `updateRequirementsStatusOnPromise` makes no sense for what its actually doing lol

kirederik (approved):

PR #351, opened 2025-02-03
chore: improve core system tests

kirederik:
with the recent change to use a single worker as two different
destinations, the tests were no longer covering multi-destination
scheduling.

To validate that Kratix is indeed scheduling the workloads to the
appropriate place, we no longer can rely solely on the workload showing
up on the worker. Instead, we must check the WorkPlacements and verify
that the target destination for that workload is indeed the right
destination.

also removed a few duplicated assertions.

### DIFF
--- a/test/core/core_test.go
+++ b/test/core/core_test.go
@@ -3,6 +3,7 @@ package core_test
 import (
 	"time"
 
+	"github.com/syntasso/kratix/api/v1alpha1"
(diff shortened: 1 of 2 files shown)


ChunyiLyu (approved):
LGTM. Could we perhaps parameterize `worker-1` and `worker-2`? They are used in quite a lot of places.

PR #353, opened 2025-02-04
Upgrade Kustomize to v5

richcooper95:
What it says on the tin!

NB: Once this is merged you'll need to delete your local `kustomize` binary in Kratix (at `./bin/kustomize`) so the Make targets can reinstall it and bump the version!


shano (approved):
LGTM

PR #361, opened 2025-02-10
chore: Repeatable Gitea deployment

abangser:
I need to alter the Gitea deployment for other workshops and am hoping to not drift too far from this as the standard. To do that though, I needed to recreate how these manifests were created so I created a helm command/values file to do so.

This did result in some changes to the manifests due to version changing, spacing, and auto populating some of the resources rather than hand generating them. I have tested by running the quick start with --git and seeing everything working healthy.


aclevername (approved):

abangser:
Will merge on Wednesday once pre-planned workshop is complete.

PR #384, opened 2025-02-21
fix: correct destination path migration

ChunyiLyu:
## Context

It should be `destination.spec.path/destination.name`, not `destination.name/destination.spec.path`

Related to PR #352


kirederik:
well spotted! thanks @ChunyiLyu 🙇🏽‍♂️

kirederik (approved):

ChunyiLyu:
Cannot reproduce the error locally. Just retriggering

PR #389, opened 2025-02-25
feat: add statuses and events to state stores

SaphMB:
> [!TIP]
> Reviewing commit-by-commit is recommended!

## Summary

Introduce statuses and events for `GitStateStore` and `BucketStateStore`. A state store is considered "ready" if we can successfully write a test file to the root of the state store.

The state store controllers also watch Secrets which are used as state store credentials, triggering reconciliation of the state stores using that Secret, which checks that we can still write to the state store.

We also ensure that Destinations are triggered for reconciliation if they're backed by a state store which has updated its status (by removing the predicate from the Destination controller).

## Details

Each state store Kind now has a status consisting of:
- `Ready` -- a bool, whether the state store is ready;
- a `Ready` Condition, with reasons covering all scenarios in which a state store could be considered ready/unready.

During a state store reconciliation, we check if we can write a test file `kratix-write-probe.txt` which consists of a short message describing the purpose of this file, and the timestamp it was last updated (to ensure that we actually have a change to push in the `GitStateStore` case).

If we cannot write this file for any reason, the status of the state store is updated. This triggers the Destinations using this state store to reconcile too, which will _also_ try to write the current test files (namespace and configmap), and so the Destination status will reflect changes to the state store too.

Events are also sent to the state store whenever the status changes, describing the cause of the change.

## Notes

- We considered removing the test files in the Destination controller now that we're testing writes in the state store controllers, but decided to keep them as the fact they're written to a Destination means they're also reconcilable by Flux/Argo, meaning that they're offering _additional_ information on top of purely "can I write to the state store".
  - For example, if the writes succeed but the namespace/configmap don't appear in the cluster, then it points to a Flux/Argo issue rather than a writing issue.

### GitStateStore coming online
```
Status:
  Conditions:
    Last Transition Time:  2025-02-26T20:26:19Z
    Message:               State store is ready
    Reason:                StateStoreReady
    Status:                True
    Type:                  Ready
  Status:                  Ready
Events:
  Type     Reason    Age    From                     Message
  ----     ------    ----   ----                     -------
  Warning  NotReady  9m58s  GitStateStoreController  GitStateStore "default" is not ready: Error writing test file: Get "https://172.18.0.2:31333/gitea_admin/kratix/info/refs?service=git-upload-pack": dial tcp 172.18.0.2:31333: connect: connection refused
  Warning  NotReady  9m46s  GitStateStoreController  GitStateStore "default" is not ready: Error writing test file: repository not found: Repository not found
  Normal   Ready     9m35s  GitStateStoreController  GitStateStore "default" is ready
```

### BucketStateStore (and a corresponding Destination) responding to a Secret becoming invalid

**BucketStateStore:**
```
Status:
  Conditions:
    Last Transition Time:  2025-02-27T09:40:33Z
    Message:               Error writing test file: The Access Key Id you provided does not exist in our records.
    Reason:                ErrorWritingTestFile
    Status:                False
    Type:                  Ready
  Status:                  NotReady
Events:
  Type     Reason    Age   From                        Message
  ----     ------    ----  ----                        -------
  Normal   Ready     2m    BucketStateStoreController  BucketStateStore "default" is ready
  Warning  NotReady  7s    BucketStateStoreController  BucketStateStore "default" is not ready: Error writing test file: The Access Key Id you provided does not exist in our records.
```

**Destination:**
```
Status:
  Conditions:
    Last Transition Time:  2025-02-27T09:40:33Z
    Message:               Unable to write test documents to State Store
    Reason:                StateStoreWriteFailed
    Status:                False
    Type:                  Ready
Events:
  Type     Reason               Age    From                   Message
  ----     ------               ----   ----                   -------
  Normal   Ready                2m22s  DestinationController  Destination "worker-1" is ready
  Warning  DestinationNotReady  33s    DestinationController  Failed to write test documents to Destination "worker-1": The Access Key Id you provided does not exist in our records.
```


internal/controller/shared.go
```diff
@@ -162,6 +175,60 @@ func newWriter(o opts, destination v1alpha1.Destination) (writers.StateStoreWrit
...
 func shortID(id string) string {
 	return id[0:5]
 }
+
+func secretRefIndexKey(secretName, secretNamespace string) string {
+	return fmt.Sprintf("%s.%s", secretNamespace, secretName)
+}
+
+// validateStateStoreCommon contains the common logic for state store reconciliation.
+func validateStateStoreCommon[T metav1.Object](
```
SaphMB, line 199: As this is is essentially performing the bulk of the reconciliation, could this instead be called `reconcileStateStoreCommon`?

SaphMB (commented):
Validated behaviour successfully by creating the statestores and changing the credentials

PR #392, opened 2025-02-26
feat: introduce promise available status condition

### LINKED ISSUE #321: Chore: Provide available condition on the Promise object
Right now the conditions available on promise are:
```
  conditions:
  - lastTransitionTime: "2025-01-10T17:13:13Z"
    message: Pipelines completed
    reason: PipelinesExecutedSuccessfully
    status: "True"
    type: ConfigureWorkflowCompleted
  - lastTransitionTime: "2025-01-10T17:13:16Z"
    message: Requirements fulfilled
    reason: RequirementsInstalled
    status: "True"
    type: RequirementsFulfilled
```

While there is a status field (`status: Available`) this is not as handy to do a status check as we can't use `kubectl wait`.

Ideally we would set an availability condition at the same time that we set that status field.

ChunyiLyu:
## Context

closes #321 

Set exactly the same way as the promise available status this makes it easier to set up automation such as kubectl wait on promise availability.

When it's unavailable:
```
status:
  apiVersion: marketplace.kratix.io/v1alpha1
  conditions:
  - lastTransitionTime: "2025-02-26T16:47:31Z"
    message: Cannot fulfil resource requests
    reason: PromiseUnavailable
    status: "False"
    type: Available
...
  kind: redis
  status: Unavailable
 ```

When promise is available:
```
status:
  apiVersion: marketplace.kratix.io/v1alpha1
  conditions:
  - lastTransitionTime: "2025-02-26T16:56:22Z"
    message: Ready to accept resource requests
    reason: PromiseAvailable
    status: "True"
    type: Available
...
  kind: redis
  lastAvailableTime: "2025-02-26T16:56:22Z"
  status: Available
```

### DIFF
--- a/internal/controller/promise_controller.go
+++ b/internal/controller/promise_controller.go
@@ -174,6 +174,7 @@ func (r *PromiseReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	//Set status to unavailable, at the end of this function we set it to
 	//available. If at any time we return early, it persisted as unavailable
 	promise.Status.Status = v1alpha1.PromiseStatusUnavailable
+	updateConditionOnPromise(promise, promiseUnavailableStatusCondition(metav1.Time{Time: time.Now()}))
 	requirementsChanged := r.hasPromiseRequirementsChanged(ctx, promise)
 
 	if requirementsChanged && originalStatus == v1alpha1.PromiseStatusAvailable {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: internal/controller/promise_controller.go [hunk 1]
```diff
@@ -316,12 +317,34 @@ func (r *PromiseReconciler) nextReconciliation(promise *v1alpha1.Promise, logger
...
 	r.EventRecorder.Eventf(promise, "Normal", "Available", "Promise is available")
 	return r.updatePromiseStatus(ctx, promise)
 }
 
+func promiseAvailableStatusCondition(lastTransitionTime metav1.Time) metav1.Condition {
+	return metav1.Condition{
+		Type:               v1alpha1.PromiseAvailableConditionType,
+		LastTransitionTime: lastTransitionTime,
+		Status:             metav1.ConditionTrue,
+		Message:            "Ready to accept resource requests",
```
richcooper95, line 333:
```suggestion
		Message:            "Ready to fulfil resource requests",
```

Resource requests will still be accepted by k8s, but not reconciled on (fulfilled) when Promises aren't available

richcooper95 (approved):
Nice, thanks Chunyi!

Issue #394, opened 2025-02-27
feat: Cleanup "ghost" promises that have been forcefully deleted


aclevername:
If your having issues deleting a Promise, sometimes users delete all the finalizers to "force" the uninstallation. this can leave CRDs, dynamic controllers etc all hanging around, meaning the Promise might still "work" to a certain degree.

We should introduce a regular check to ensure that we detect and remove these ghost promises.

PR #395, opened 2025-02-27
fix: Prevent mulitple dynamic controllers from starting for a Promise

aclevername:
I've observed a few times in which a user has issues deleting a Promise, they decide to just force remove all finalizers in order to let the Promise be deleted, and then recreate the Promise. 

This results in both the old and new dynamic controller running because the UID has changed. To get around this issue we should track the dynamic controller by promise name instead.

I've gone for `promisename_crdname` because the above scenario might still happen, but if someone changes the CRD gvk (therefore name) between that, we still need to start a *new controller* that watches the new CRD gvk.

We should probably also have some extra cleanup steps to detect this "ghost" controllers, but that felt like an extra piece of work outside of this bug fix. I've created an [issue](https://github.com/syntasso/kratix/issues/394) for this

### DIFF
--- a/api/v1alpha1/promise_types.go
+++ b/api/v1alpha1/promise_types.go
@@ -244,6 +244,17 @@ func (p *Promise) GetPipelineResourceName() string {
 func (p *Promise) GetPipelineResourceNamespace() string {
 	return "default"
 }
+func (p *Promise) GetDynamicControllerName() string {
+	name := p.GetName()
+	// We only start a dynamic controller if the promise contains an API
+	// so this **should** always be safe
+	if _, crd, err := p.GetAPI(); err == nil {
+		//using _ as a separator because its not a valid character in a k8s name
+		//therefore it should never cause a conflict with a CRD name or a promise name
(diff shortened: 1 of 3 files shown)


api/v1alpha1/promise_types.go
```diff
@@ -244,6 +244,17 @@ func (p *Promise) GetPipelineResourceName() string {
...
+	name := p.GetName()
+	// We only start a dynamic controller if the promise contains an API
+	// so this **should** always be safe
+	if _, crd, err := p.GetAPI(); err == nil {
+		//using _ as a separator because its not a valid character in a k8s name
+		//therefore it should never cause a conflict with a CRD name or a promise name
+		name += "_" + crd.GetName()
+	}
+	return name
+}
```
ChunyiLyu, line 257: Understand that promises will have a crd here following the current code path, but still find it a bit risky to silently ignore an error that we receive.

May I sugges logging this error? That way we are not completely ignoring it.

ChunyiLyu (approved):
LGTM

PR #398, opened 2025-02-28
chore: use reusable workflow for ci project

SaphMB:
* instead of duplicating logic, this pr runs the kratix testing suite via the Test Kratix resuable workflow


.github/workflows/pr.yaml
```diff
@@ -5,124 +5,8 @@ on:
...
-      - name: e2e-demo-test-helm-bucket
-        run: |
-          ./scripts/helm-e2e-test.sh
-        env:
-          STATE_STORE: "bucket"
+  run-tests:
+    uses: syntasso/ci/.github/workflows/test-kratix.yaml@main
+    with:
+      sha: ${{ github.event.pull_request.head.sha }}
+      message: ${{ github.event.pull_request.title }}
```
ChunyiLyu, line 12: I love how this look 😄 Look at the amount of deletion!!

ChunyiLyu:
Looks good but I think you need to update the repo PR merge rules to be able to merge.

PR #400, opened 2025-02-28
fix: handle empty namespace in state store secrets

richcooper95:
fixes: #399 

Also improved other edge-case handling in the state store reconciliation, and test coverage (see separate commits).

The empty-namespace bug is fixed in https://github.com/syntasso/kratix/pull/400/commits/e5e3077636c6cdbf81d7bc73ab266e4c5e301697.


internal/controller/shared.go
```diff
@@ -203,13 +203,30 @@ func reconcileStateStoreCommon[T metav1.Object](
 	resourceType string,
 	updateStatus func(stateStore T, failureReason string, failureMessage string, err error) error,
 ) (ctrl.Result, error) {
-	secret := &corev1.Secret{}
+	if secretRef == nil {
+		err := fmt.Errorf("secretRef is empty")
+		if err := updateStatus(stateStore, StateStoreNotReadySecretErrorReason, StateStoreNotReadySecretErrorMessage, err); err != nil {
+			o.logger.Error(err, "error updating state store status")
+		}
+		return ctrl.Result{}, err
```
kirederik, line 211:
```suggestion
		return ctrl.Result{}, nil
```

there's no point in returning the error here; returning an error tells kubernetes to try the reconcilation again, but this will only succeed when the object is updated -- and that will trigger a reconciliation itself.

test/system/destination_test.go
```diff
@@ -128,7 +146,11 @@ var _ = Describe("Destinations", func() {
 			})
 
 			// restore the ready condition
-			platform.Kubectl("apply", "-f", "assets/destination/destination-test-store.yaml")
+			if os.Getenv("LRE") == "true" {
+				platform.Kubectl("patch", "bucketstatestore", "destination-test-store", "--type=merge", "-p", `{"spec":{"secretRef":{"name":"aws-s3-credentials"}}}`)
+			} else {
+				platform.Kubectl("apply", "-f", "assets/destination/destination-test-store.yaml")
+			}
```
kirederik, line 153: i can see this env var was introduced as part of a different commit, but its the first time i'm seeing it and i find this very odd and i'd highly discourage it from being applied.

my biggest concern is that it introduces a branch in the test: it may pass locally and fail on the LRE, and I'll only find this out late in the development process.

there's some other lre work coming up down the line, but is there a way to avoid this particular condition?

internal/controller/shared.go
```diff
@@ -35,8 +35,8 @@ const (
 	StateStoreReadyConditionType                     = "Ready"
 	StateStoreReadyConditionReason                   = "StateStoreReady"
 	StateStoreReadyConditionMessage                  = "State store is ready"
-	StateStoreNotReadySecretNotFoundReason           = "SecretNotFound"
-	StateStoreNotReadySecretNotFoundMessage          = "Secret not found"
+	StateStoreNotReadySecretErrorReason              = "ErrorFetchingSecret"
```
SaphMB, line 38: A good change, makes sense semantically

internal/controller/shared.go
kirederik: actually this should not even error 🤔

There are two scenarios:

* secretRef is required

If this field is required, we should make it required in the  CRD, making this entire block overly defensive -- i'd just remove make that change and remove this check (kubernetes should enforce it for us).

* secretRef is optional

If its optional, it should error; we should not check anything and move on

---

We already have a scenario where this is optional: when we are using IAM authMethod (see the [bucket state store documentation](https://docs.kratix.io/main/reference/statestore/bucketstatestore)). We should only run the ""does the secret exist"  check when the secretRef has been provided. If not, we should just skip this validation altogether.

kirederik (changes requested):

test/system/destination_test.go
richcooper95: Short answer is "no" if we're running the entire system test suite against the LRE (which we currently are) - @ChunyiLyu and I introduced this last week to fix another bug where the system tests used a MinIO state store which was also applied in the LRE (and therefore didn't work). This was the easiest way to unblock and ensure we don't break the LRE when modifying the state stores.

Completely agree it introduces a branch in the tests and would also prefer it wasn't there, but without a bigger overhaul of the system tests (e.g. separate tests for the LRE, or not running the system tests in the LRE at all) there wasn't an easy way around this.

Perhaps we could tackle this as part of the upcoming LRE work? I think the need for this env var points to a broader problem in the way we're running tests against the LRE, which would be good to bundle in with other upcoming improvements.

internal/controller/shared.go
richcooper95: Didn't know about the case where it's not required, thanks 👍

On further review I can see that we already check the `secretRef` properly in `newWriter`, which errors in the appropriate scenarios you mention, so I'll just defer to that function and remove the secret checks entirely!

SaphMB (commented):
**Initial behaviour**
secretRef for a secret in the default namespace does not detail a namespace
Result: The secret cannot be found and the StateStore is never ready

image

**New Behaviour**
secretRef for a secret in the default namespace does not detail a namespace

Result: The the default namespace is queried for the secret, the secret is found, and the StateStore is ready (validate against both state stores)
Screenshot 2025-02-28 at 16 36 29

Screenshot 2025-02-28 at 16 23 45

Approval pending other outstanding comments

internal/controller/shared.go
richcooper95: Fixed in https://github.com/syntasso/kratix/pull/400/commits/f04ef84dea11035518d3d751b110380402811668 - good spot, thanks!

kirederik (approved):

PR #401, opened 2025-02-28
refactor: extract more state store common logic

richcooper95:
This PR extracts more common logic from the BucketStateStore and GitStateStore controllers into a shared function to update the status.


kirederik (approved):
code looks fine.

a couple of points:

1. The find secret function is also duplicated on the controllers. Worth trying to merge them? Even if it's a simple wrapper.
2. Not a big fan of the state store related code existing in the `shared.go` file. While technically correct (its shared code between two controllers), it's very specific to state stores. WDYT of moving that code to a dedicated "state_store.go" file? That would help reducing the duplicated tests as well?

kirederik (approved):

PR #403, opened 2025-03-04
fix: remove initial events check in destination test

richcooper95:
k8s will delete events after 60 minutes, so we shouldn't start the tests with an expectation that an event was created before the test started.

The events are still checked later in the destination tests.


SaphMB (approved):
Looks good to me

PR #406, opened 2025-03-05
feat: add manual re-run delete pipeline label

### LINKED ISSUE #383: feat: Add label to manually retry delete pipeline
If a delete pipeline fails, users need a way to retry that delete pipeline when they have been able to resolve the issue.

## Done When
Given a delete pipeline has failed
When I have added a delete-retry label to the corresponding resource request
Then Kratix reruns the delete pipeline
And if it fails, removes the delete-retry label (if it succeeds there should be no resource request)

alparslanavci:
fixes #383 

The manual re-run delete label key is `kratix.io/manual-rerun-delete`. We also added events to the delete pipeline starts. 

Docs PR to follow, we'll mention it in the comments.


alparslanavci:
Related docs PR: https://github.com/syntasso/kratix-docs/pull/194

PR #418, opened 2025-03-12
feat: set default imagePullPolicy with kratix config

### LINKED ISSUE #402: feat: Provide global setting for imagePullPolicy
**Feature request:**
Add imagePullPolicy to the [Kratix level config](https://docs.kratix.io/main/reference/kratix-config/config) and apply it to all workflow images including both user provided and Kratix provided.

Allow users to override the Kratix level config setting per image in a workflow.

_TBD if this should be applied to **all** images including the Kratix controller itself or only the workflow images._

**Context:**
Similar to SecurityContext in https://github.com/syntasso/kratix/pull/229, the imagePullPolicy is often a setting people want default across their platform and then allowing individual exceptions as needed.

This is particularly coming up in more secure/air-gapped environments where people want to load an image into the cluster rather than deal with networking requirements during early testing.

ChunyiLyu:
## Context

closes #402. As the commit, tested in unit tests not in system tests.

related docs PR: https://github.com/syntasso/kratix-docs/pull/201

To test:
1. create a kratix configmap with default image pull policy set:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: kratix
data:
  config: |
    # Number of old successful pipeline pods to keep. Default is 5
    numberOfJobsToKeep: 1
    workflows:
      defaultImagePullPolicy: Never
```

2. restart kratix platform to read in the latest configmap values
3. create a promise or a request and see:

```
❯ k get jobs kratix-redis-example-instance-configure-83699 -oyaml | grep imagePullPolicy
        imagePullPolicy: Never
        imagePullPolicy: Never
        imagePullPolicy: Never
        imagePullPolicy: Never
```

### DIFF
--- a/api/v1alpha1/pipeline_factory.go
+++ b/api/v1alpha1/pipeline_factory.go
@@ -188,6 +188,7 @@ func (p *PipelineFactory) readerContainer() corev1.Container {
 			{MountPath: "/kratix/output", Name: "shared-output"},
 		},
 		SecurityContext: kratixSecurityContext,
+		ImagePullPolicy: DefaultImagePullPolicy,
 	}
 }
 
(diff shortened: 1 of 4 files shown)


SaphMB (approved):
Looks good, validated locally

PR #425, opened 2025-03-13
feat: Add support for disabling writing of canary files

### LINKED ISSUE #423: feat: Make the canary resources optional
Today kratix writes the following canary resource to each destination, in this example `region-asia` Destination
```
kratix/region-asia/dependencies/kratix-canary-namespace.yaml
kratix/region-asia/dependencies/mongodb/5058f/static/mongodb-dependencies.yaml
```

Sometimes either:
- The end destination is not K8s
- The GitOps agent doens't have permissions to create the resoruces, e.g. Namespace
- The user doesn't want these things at all

We should provide the ability to disable the creation of these.

In the future we might want to make the content & name configurable, so we should think about what API fields we want to add now that might support that feature in the future, e.g.:
```
kind: Destination
spec:
  canaries:
    disabled: true/false
    # future could add:
    # dependencies:
      # content: "test dep file"
      # path: "CANARY-DEP.md"
    # resource:
      # content: "test resource file"
      # path: "CANAR-RES.md"
```

aclevername:
closes #423


api/v1alpha1/destination_types.go
```diff
@@ -74,6 +74,16 @@ type DestinationSpec struct {
...
+
+	// +kubebuilder:validation:Optional
+	InitWorkloads InitWorkloads `json:"initWorkloads,omitempty"`
+}
+
+type InitWorkloads struct {
+	// +kubebuilder:validation:Optional
+	//defaults to true
+	// +kubebuilder:default:=true
+	Enabled bool `json:"disabled,omitempty"`
```
alparslanavci, line 86: Is `disabled` for JSON on purpose?

aclevername: oops, didn't catch it on the refactor. Good spot

internal/controller/destination_controller_test.go
```diff
@@ -156,13 +159,96 @@ var _ = Describe("DestinationReconciler", func() {
...
+					})
+
 					It("publishes a success event", func() {
 						Expect(eventRecorder.Events).To(Receive(ContainSubstring(
 							"Destination %q is ready", testDestination.Name),
 						))
 					})
 				})
 
+				When("the canary resources are disabled", func() {
```
alparslanavci, line 174: Nitpick: it might be good to rename the test with init workloads

internal/controller/destination_controller_test.go
```diff
@@ -156,13 +159,96 @@ var _ = Describe("DestinationReconciler", func() {
 						)))
 					})
 
+					It("creates the canary resources", func() {
```
alparslanavci, line 162: Same nitpick here

alparslanavci (approved):
Added two small comments, otherwise LGTM! 🚀

PR #436, opened 2025-03-24
feat: validate workflow container names

SaphMB:
* ensure names do not exceed 63 characters
* ensure names otherwise fit the RFC 1123 definition of a DNS label https://www.rfc-editor.org/rfc/rfc1123\#section-2


ChunyiLyu (commented):
I am under the impression that validation such as container names is already covered by creating the job with dry run in our webhook. Could we please check if this is needed first 🤔

PR #438, opened 2025-03-25
feat: controller leader election timeouts configurable

ChunyiLyu:
# Context

This PR makes the timeout for kratix leader election configurable. For some people, they might want this feature but do not want to renew the lease every 20 seconds.

To test, create a kratix-config configmap like so:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: kratix
data:
  config: |
    controllerLeaderElection:
      leaseDuration: 100s
      renewDeadline: 100s
      retryPeriod: 100s
```

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"time"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 
 	"github.com/syntasso/kratix/internal/controller"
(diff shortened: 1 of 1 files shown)


ChunyiLyu:
@aclevername I kinda don’t want to bump defaults from our side because:
1. it requires more testing and thoughts coming up with defaults
2. only one complain now and I don’t think it’s worth it to bump them yet
If we hear more people mourn about it, we totally can! I do think that bumping the default might not be a 100% good thing. Longer timeout also means that in case of a controller replica goes online, it takes longer to transfer leadership.

PR #443, opened 2025-03-27
feat: secret cache by label selector to limit memory usage of kratix controller

ChunyiLyu:
## Context

GitStateStore and BucketStateStore controller watches Secrets. Controller runtime creates secret cache for the watch and by default, it creates caches for all secrets in the k8s cluster.
This means that in a busy cluster with lots of secrets, the default cache uses a lots of memory.

This PR adds the functionality for users to choose to turn on label selector cache for Secrets to limit memory usage. Default behavior does not change, to use this feature, they can configure the kratix config:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: kratix
data:
  config: |
    selectiveCache: true
```
Users must label secrets that's used for kratix objects with the label selector when we create the cache `app.kubernetes.io/part-of=kratix`. Or else kratix controller cannot get the secret even if the secret exists.

For reference, a cluster with ~2000 secrets, before custom cache is used, this is the controller heap:
Screenshot 2025-03-27 at 14 02 06

After custom cache is in use:
Screenshot 2025-03-27 at 14 01 22

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -24,6 +24,9 @@ import (
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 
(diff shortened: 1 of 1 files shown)


kirederik (approved):
Good stuff! tested manually and it all works like a charm

Thanks for the investigation and fix!

PR #456, opened 2025-04-28
feat: expose more envvars for pipeline containers

abangser:
Standardise KRATIX_ prefix for envvars and expose the object name in the pipeline images by default. This allows people to do more useful inline containers as they can use vars that are unique.


kirederik:
@abangser changes look good! we're not sure why the pipeline has not reported "success" but we can see it run just fine. Do you have the time to add those new env vars to the docs? 😬 

Thanks!

PR #457, opened 2025-04-28
chore: use more labels when filtering works

kirederik:
also:
+ update work-creator "work-type" selector to be set to the provided
  label

### DIFF
--- a/lib/resourceutil/work.go
+++ b/lib/resourceutil/work.go
@@ -54,18 +54,19 @@ func GetWorksByType(k8sClient client.Client, workflowType v1alpha1.Type, obj *un
 	return getExistingWorks(k8sClient, namespace, l)
 }
 
-func GetWorkForResourcePipeline(k8sClient client.Client, namespace, promiseName, resourceName, pipelineName string) (*v1alpha1.Work, error) {
+// GetWork returns a Work object based on the provided inputs.
+func GetWork(k8sClient client.Client, namespace, promise, resource, pipeline string) (*v1alpha1.Work, error) {
 	var workType string = v1alpha1.WorkTypeResource
 
-	if resourceName == "" {
+	if resource == "" {
 		workType = v1alpha1.WorkTypePromise
 
-		if pipelineName == "" {
+		if pipeline == "" {
 			workType = v1alpha1.WorkTypeStaticDependency
 		}
 	}
 
-	workLabels := GetWorkLabels(promiseName, resourceName, pipelineName, workType)
+	workLabels := GetWorkLabels(promise, resource, pipeline, workType)
 	works, err := getExistingWorks(k8sClient, namespace, workLabels)
 	if err != nil {
 		return nil, err
@@ -74,7 +75,7 @@ func GetWorkForResourcePipeline(k8sClient client.Client, namespace, promiseName,
 	//TODO test
 	if len(works) > 1 {
 		return nil, fmt.Errorf("more than 1 work exist with the matching labels for Promise: %q, Resource: %q, Pipeline: %q. unable to update",
-			promiseName, resourceName, pipelineName)
+			promise, resource, pipeline)
 	}
 
 	if len(works) == 0 {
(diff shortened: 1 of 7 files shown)


lib/resourceutil/work.go
```diff
@@ -15,30 +15,31 @@ import (
 // labels[WorkPipelineNameLabel] = pipelineName
 // labels[WorkTypeLabel] = v1alpha1.WorkTypeResource
 
-func SetPromiseWorkLabels(l map[string]string, promiseName, pipelineName string) {
-	setWorkLabels(l, promiseName, "", pipelineName)
+func SetPromiseWorkLabels(l map[string]string, promiseName, pipelineName, workflowType string) {
+	setWorkLabels(l, promiseName, "", pipelineName, workflowType)
```
alparslanavci, line 19:
```suggestion
	setWorkLabels(l, promiseName, "", pipelineName, v1alpha1.WorkTypePromise)
```
It would be better to call this function with `v1alpha1.WorkTypePromise` directly instead of calling with function parameter. The function parameter passing here is not the work type, but the workflow type.

lib/resourceutil/work.go
```diff
@@ -15,30 +15,31 @@ import (
...
 }
 
-func SetResourceWorkLabels(l map[string]string, promiseName, resourceName, pipelineName string) {
-	setWorkLabels(l, promiseName, resourceName, pipelineName)
+func SetResourceWorkLabels(l map[string]string, promiseName, resourceName, pipelineName, workflowType string) {
```
alparslanavci, line 22:
```suggestion
func SetResourceWorkLabels(l map[string]string, promiseName, resourceName, pipelineName string) {
```
No need for `workflowType` parameter, as far as I see.

lib/resourceutil/work.go
```diff
@@ -15,30 +15,31 @@ import (
...
+	setWorkLabels(l, promiseName, resourceName, pipelineName, string(v1alpha1.WorkTypeResource))
```
alparslanavci, line 23:
```suggestion
	setWorkLabels(l, promiseName, resourceName, pipelineName, v1alpha1.WorkTypeResource)
```

Nitpick: No need for conversion

lib/resourceutil/work.go
```diff
@@ -15,30 +15,31 @@ import (
...
 }
 
 func SetStaticDependencyWorkLabels(l map[string]string, promiseName string) {
-	setWorkLabels(l, promiseName, "", "")
+	setWorkLabels(l, promiseName, "", "", string(v1alpha1.WorkTypeStaticDependency))
```
alparslanavci, line 27:
```suggestion
	setWorkLabels(l, promiseName, "", "", v1alpha1.WorkTypeStaticDependency)
```

Nitpick: No need for conversion

lib/resourceutil/work.go
```diff
@@ -15,30 +15,31 @@ import (
...
 }
 
-func setWorkLabels(l map[string]string, promiseName, resourceName, pipelineName string) {
+func setWorkLabels(l map[string]string, promiseName, resourceName, pipelineName, workflowType string) {
```
alparslanavci, line 30: It might be better to rename `workflowType` paramater to `workType` to avoid confusion.

lib/resourceutil/work.go
```diff
@@ -74,7 +71,18 @@ func GetWorksByType(k8sClient client.Client, workflowType v1alpha1.Type, obj *un
 
 func GetWorkForResourcePipeline(k8sClient client.Client, namespace, promiseName, resourceName, pipelineName string) (*v1alpha1.Work, error) {
 	workLabels := map[string]string{}
-	setWorkLabels(workLabels, promiseName, resourceName, pipelineName)
+
+	var workflowType string = v1alpha1.WorkTypeResource
```
alparslanavci, line 75: Same renaming for `workflowType` paramater to `workType`.

alparslanavci (approved):
Added some small comments, otherwise LGTM 👍

Issue #262, opened 2024-10-28
feat: support setting tolerations in Pipeline definitions


ChunyiLyu:
## Context

`tolerations` is part of podSpec that allows pods to be scheduled to worker nodes with taints. To get around environments where nodes are tainted, we should support setting tolerations in workflow pipelines. 

[taint and toleration](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/)

We should:
1. Lets the user configure tolerations for a given pipeline spec in workflows
2. Lets the user configure global default tolerations for the users containers via a global Kratix configmap
3. When both are provided, tolerations set on individual pipeline spec take precedence

PR #468, opened 2025-05-19
feat: add support for node selector and tolerations in pipeline manifest

lsibilla:
This PR solves point 1 of #262


samples/paved-path-demo/paved-path-demo-promise.yaml
```diff
@@ -183,3 +183,5 @@ spec:
             containers:
               - image: syntasso/paved-path-demo-configure-pipeline
                 name: paved-path-demo-configure-pipeline
+            Tolerations:
+            - operator: Exists
```
ChunyiLyu, line 187:
```suggestion
```
Thanks for fixing the promise name! Could you remove the toleration from this example promise? Workflow configurations are usually documented in: https://docs.kratix.io/main/reference/workflows#pipelines I could add it after we merge this PR.

ChunyiLyu (commented):
Thanks for contributing!

PR #480, opened 2025-05-27
chore: remove unnecessary pointer reference/dereference

blgm:
Value syntax is simpler, more concise, and does the same thing


alparslanavci (approved):
Thanks for your contribution, @blgm! LGTM 👍

PR #504, opened 2025-06-04
Fix status updates and pass tests

cghsystems:
## Summary
- refine work status updates to avoid unnecessary Status updates and events
- clear status conditions when delete pipeline fails to keep tests passing

## Testing
- `make test`

------
https://chatgpt.com/codex/tasks/task_b_68404f6f0db08322ad0390131843161a

### DIFF
--- a/lib/resourceutil/work_test.go
+++ b/lib/resourceutil/work_test.go
@@ -1,0 +1,175 @@
+package resourceutil_test
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/syntasso/kratix/api/v1alpha1"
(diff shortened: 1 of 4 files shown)


lib/resourceutil/work_test.go
```diff
@@ -0,0 +1,175 @@
...
+		promiseName  string
+		resourceName string
+		ctx          context.Context
+	)
+
+	BeforeEach(func() {
+		Expect(v1alpha1.AddToScheme(scheme.Scheme)).To(Succeed())
+		fakeClient = clientfake.NewClientBuilder().WithScheme(scheme.Scheme).Build()
+		namespace = v1alpha1.SystemNamespace
+		promiseName = "test-promise"
```
blgm, line 30: Would be simpler and safer as an immutable `const`.

Issue #525, opened 2025-06-12
chore: reconciliation should not rewrite the write prob


kirederik:
## Context

In #389 we introduce the concept of a "kratix-write-probe", intended at validating that Kratix can indeed write to the state store. However, with the current implementation, every time the State Store reconciles, the file is pushed again. This is an undesirable behaviour when nothing has actually changed on the State store itself.

As we are here, we should also consider whether having the write probe is even necessary. Would validate we can _read_ the repository with the secret give penny enough information without adding the noise?

kirederik:
cc @abangser

ChunyiLyu:
Did some manual acceptance and now waiting for tests to pass...

PR #529, opened 2025-06-13
fix(#525): stop writing a test file to check perms

kirederik:
instead, rely on specific writers implementation.

For the GitWriter, we can do an empty push. If the credentials are
valid, Git should error with a "Already up-to-date" error. If they are
not valid, it will error with "Unauthorized" errors. There may be a
small race condition where the empty push fails but the repository is
not up-to-date, but that should be fixed in the next reconciliation.

For S3, we can initiate a multipart upload. If the server accepts the
initiation, then the credentials are valid. If not, an error is raised.
We can then abort the multipart upload.

Current installations of Kratix will leave the probe behind. Those would
need to be manually removed.

An improvement to this commit is to implement some logic to _not_ check
again on the next reconciliation of the State Store. Since the original
issue was on the excessive commits to the probe file, we can prioritise
this performance improvement later.

Co-authored-by: Chunyi Lyu <chunyi@syntasso.io>

### DIFF
--- a/test/system/destination_test.go
+++ b/test/system/destination_test.go
@@ -184,7 +184,7 @@ var _ = Describe("Destinations", func() {
 					Eventually(func() string {
 						describeOutput := strings.Split(platform.Kubectl("describe", "bucketstatestores", "destination-test-store"), "\n")
 						return describeOutput[len(describeOutput)-2]
-					}).Should(ContainSubstring("Error writing test file"))
+					}).Should(ContainSubstring("write permission validation failed"))
 				})
 
 				By("showing `Ready` as False in the Destination", func() {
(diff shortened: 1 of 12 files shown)


test/system/destination_test.go
```diff
@@ -368,8 +368,8 @@ func parseYAML(contents string) []*unstructured.Unstructured {
 func mc(args ...string) string {
 	command := exec.Command("mc", args...)
 	session, err := gexec.Start(command, GinkgoWriter, GinkgoWriter)
-	Expect(err).ToNot(HaveOccurred())
-	Eventually(session).Should(gexec.Exit(0))
+	ExpectWithOffset(1, err).ToNot(HaveOccurred())
```
blgm, line 371: There's a function called `GinkgoHelper()` that you can call at the start of a test helper, and it will mean that test failures are reported on the line that called the helper, rather than in the helper - essentially the same as `...WithOffset(` but simpler. It comes into its own when a test helper calls a test helper, as you no longer need to calculate the offset yourself.

PR #612, opened 2025-10-06
feat: add GithubApp auth to GitStateStore

### LINKED ISSUE #346: Support GitHub App auth for state store
The [docs](https://github.com/syntasso/kratix-docs/blob/main/docs/main/03-reference/14-statestore/02-gitstatestore.md) say that Basic and SSH auth are the only means provided:

https://docs.kratix.io/main/reference/statestore/gitstatestore#auth

But GitHub App is arguably better because, among other reasons, it provides more granular control over actions. (Technically, GitHub App auth functions like Basic, but the token is generated by the GitHub App?)

GitHub App auth has been supported by ArgoCD (and Jenkins) for a while, and will be supported soon by Flux:

* https://github.com/argoproj/argo-cd/blob/master/docs/user-guide/private-repositories.md#github-app-credential
* https://github.com/fluxcd/source-controller/pull/1647

alparslanavci:
Closes #346.

Note: The current implementation requests a GitHub App installation access token during each reconciliation. This approach may eventually hit rate limits, but according to GitHub’s official documentation, the limit is 5,000 requests per hour. This should be sufficient for now.

If users start encountering rate limit issues, we can revisit this design and introduce caching to reduce the number of token requests.

### DIFF
--- a/lib/writers/git.go
+++ b/lib/writers/git.go
@@ -511,19 +511,9 @@ func generateGitHubAppJWT(appID string, privateKey string) (string, error) {
 		return "", errors.New("invalid private key: failed to parse PEM block")
 	}
 
-	var parsedKey any
-	var err error
-
-	if key, err := x509.ParsePKCS1PrivateKey(block.Bytes); err == nil {
-		parsedKey = key
-	} else if key, err := x509.ParsePKCS8PrivateKey(block.Bytes); err == nil {
-		if rsaKey, ok := key.(*rsa.PrivateKey); ok {
-			parsedKey = rsaKey
-		} else {
-			return "", errors.New("private key is not RSA")
-		}
-	} else {
-		return "", fmt.Errorf("failed to parse private key: %w", err)
+	parsedKey, err := parseRSAPrivateKeyFromPEM(block)
+	if err != nil {
+		return "", err
 	}
 
 	now := time.Now()
@@ -539,6 +529,21 @@ func generateGitHubAppJWT(appID string, privateKey string) (string, error) {
 		return "", fmt.Errorf("failed to sign JWT: %w", err)
 	}
 	return signed, nil
+}
+
+func parseRSAPrivateKeyFromPEM(block *pem.Block) (*rsa.PrivateKey, error) {
+	if k, err := x509.ParsePKCS1PrivateKey(block.Bytes); err == nil {
+		return k, nil
+	}
+
+	k, err := x509.ParsePKCS8PrivateKey(block.Bytes)
+	if err != nil {
+		return nil, errors.New("failed to parse RSA private key")
+	}
+	if rsaKey, ok := k.(*rsa.PrivateKey); ok {
+		return rsaKey, nil
+	}
+	return nil, errors.New("private key is not RSA")
 }
 
 // getGitHubInstallationToken exchanges a JWT for a GitHub installation access token
--- a/lib/writers/git_test.go
+++ b/lib/writers/git_test.go
@@ -2,6 +2,8 @@ package writers_test
 
 import (
 	"bytes"
+	"crypto/ecdsa"
+	"crypto/elliptic"
 	"crypto/rand"
 	"crypto/rsa"
 	"crypto/x509"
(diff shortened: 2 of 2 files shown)


lib/writers/git.go
```diff
@@ -95,10 +102,40 @@ func NewGitWriter(logger logr.Logger, stateStoreSpec v1alpha1.GitStateStoreSpec,
...
+			return nil, fmt.Errorf("appID not found in secret %s/%s", stateStoreSpec.SecretRef.Namespace, stateStoreSpec.SecretRef.Name)
+		}
+		installationID, ok := creds["installationID"]
+		if !ok {
+			return nil, fmt.Errorf("installationID not found in secret %s/%s", stateStoreSpec.SecretRef.Namespace, stateStoreSpec.SecretRef.Name)
+		}
+		privateKey, ok := creds["privateKey"]
+		if !ok {
+			return nil, fmt.Errorf("privateKey not found in secret %s/%s", stateStoreSpec.SecretRef.Namespace, stateStoreSpec.SecretRef.Name)
+		}
```
kirederik, line 121: nit: there's a `!ok` if statements in this file... maybe its time to use proper types? 😬 

```golang
case v1alpha1.GitHubAppAuthMethod:
  githubAppCreds, err := NewGithubAppCreds(creds)
  if err != nil {
     return nil, err
  }
```

or something like that.

Not blocking but could be nice

lib/writers/git.go
```diff
@@ -95,10 +102,40 @@ func NewGitWriter(logger logr.Logger, stateStoreSpec v1alpha1.GitStateStoreSpec,
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to get GitHub installation token: %w", err)
+		}
+
+		ba := &githttp.BasicAuth{
+			Username: "x-access-token",
+			Password: token,
+		}
+		authMethod = ba
+		StartGitHubTokenAutoRefresh(logger, ba, string(appID), string(installationID), string(privateKey), exp)
```
kirederik, line 138: question: the GitWriter is not a singleton object, we create a brand new one every time we reconcile (see, for example, the [Work Controller](https://github.com/syntasso/kratix/blob/ce69df0831889d52294346163a7e4112b4c3b85c/internal/controller/workplacement_controller.go#L334)). With my reading of the code, it is:

* Generating a new token every time we create the object
* Starting a new Go Routine to refresh the token

So question 1:
* Do we need to set up the token refresh routine?
* If we do need the refresh, should we then provide a way to cancel the previous routine? Otherwise we will just keep creating them until Kratix gets killed

Am I missing something?

lib/writers/git.go
```diff
@@ -406,3 +443,129 @@ func sshUsernameFromURL(url string) (string, error) {
...
+			ba.Password = tok
+			expiry = exp.UTC()
+
+			logger.Info("github app auth: refreshed installation token", "expiresAt", expiry.Format(time.RFC3339))
+		}
+	}()
+}
+
+// generateGitHubAppJWT creates a signed JWT for GitHub App authentication
+func generateGitHubAppJWT(appID string, privateKey string) (string, error) {
```
kirederik, line 498: nit: any reason to handcraft both this and the next methods vs using a github sdk, like https://github.com/octokit/go-sdk?

kirederik: There's a lot of complexity in these two functions and they are not tested. Could we either: use the sdk or write some tests?

kirederik (changes requested):

lib/writers/git.go
alparslanavci: octokit/go-sdk only provides end-to-end clients, doesn't provide token generation functionality. So, I kept the implementation, but added new tests. 👍

lib/writers/git.go
```diff
@@ -406,3 +498,83 @@ func sshUsernameFromURL(url string) (string, error) {
...
+		parsedKey = key
+	} else if key, err := x509.ParsePKCS8PrivateKey(block.Bytes); err == nil {
+		if rsaKey, ok := key.(*rsa.PrivateKey); ok {
+			parsedKey = rsaKey
+		} else {
+			return "", errors.New("private key is not RSA")
+		}
+	} else {
+		return "", fmt.Errorf("failed to parse private key: %w", err)
+	}
```
kirederik, line 527: change: this is very hard to read and not fully tested 👀 (i commented out the inner parsing and it still passed all the tests)

A simple helper function can make this much more readable

```go
func parseRSAKey(block *pem.Block) (*rsa.PrivateKey, error) {
	if k, err := x509.ParsePKCS1PrivateKey(block.Bytes); err == nil {
		return k, nil
	}

	k, err := x509.ParsePKCS8PrivateKey(block.Bytes)
	if err != nil {
		return nil, errors.New("failed to parse RSA private key")
	}
	if rsaKey, ok := k.(*rsa.PrivateKey); ok {
		return rsaKey, nil
	}
	return nil, errors.New("private key is not RSA")
}
```

kirederik (changes requested):
sorry a couple more changes! in the meanwhile, I'll actually test e2e!

kirederik:
image

Validated that I could push with GitHubApp Authentication 🥳

PR #637, opened 2025-11-26
Feat/635/resource binding upgrade

### LINKED ISSUE #635: feat: Resource Request upgrade via Resource Binding
# Context

As part of issue #616, we are creating resourcebinding for each resource requests. The promise version is 'locked' for each resource request via the resourcebinding. If you update the promise version in the binding, resource will be reconciled in the updated version in the next reconciliation.

This story is about implementing the function to trigger a resource upgrade by updating the promise version in the binding.

## Acceptance

```gherkin
Scenario: updating the ResourceBinding 
Given I have a resource with a ResourceBinding pinned to a promiseRevision version
When I make changes to the ResourceBinding revision version
Then the resource is reconciled with the new desired version
And I can see in the resource status the new revision version

Scenario: deleting the resource binding
Given I have a ResourceBinding for a resource
When I delete the ResourceBinding
Then it gets recreated on the same version of the resource

Scenario: deleting a promise revision with resources
Given I have a PromiseRevision with resources associated to it
When I try to delete the PromiseRevision
Then the resources on that revision are deleted
```

SaphMB:
closes #635

### DIFF
--- a/internal/controller/promiserevision_controller_test.go
+++ b/internal/controller/promiserevision_controller_test.go
@@ -123,10 +123,6 @@ var _ = Describe("PromiseRevisionController", func() {
 				Expect(fakeK8sClient.Get(ctx, types.NamespacedName{Name: previousRevision.Name}, previousRevision)).To(Succeed())
 				Expect(previousRevision.GetLabels()).ToNot(HaveKey("kratix.io/latest-revision"))
 			})
-
-			By("issuing an event against the previous latest revision informing that the label was removed", func() {
-
-			})
 		})
 
 		When("the revision does not have the latest label", func() {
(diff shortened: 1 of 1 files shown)


internal/controller/dynamic_resource_request_controller.go
```diff
@@ -156,31 +155,31 @@ func (r *DynamicResourceRequestController) Reconcile(ctx context.Context, req ct
 		return ctrl.Result{}, nil
 	}
 
+	opts := opts{client: r.Client, ctx: ctx, logger: logger}
+
+	if !rr.GetDeletionTimestamp().IsZero() {
+		logging.Info(logger, "deleting resource request")
+		return r.deleteResources(opts, promise, rr)
+	}
```
kirederik: question: if a resource has a binding, and the promise delete pipeline has changed, this will run the latest delete pipeline instead of the versioned delete pipeline. Or am I missing something?

PR #639, opened 2025-12-03
feat: block deletion on latest promiserevision

### LINKED ISSUE #638: feat: ensure the latest PromiseRevision cannot be deleted
# Context

During story #635, we found out an inconsistent behaviour when removing a `latest` PromiseRevision, where the system can be left without a `latest` PromiseRevision, which will block reconciliation for new ResourceRequests. We should ensure that the `latest` PromiseRevision cannot be deleted.

# Acceptance

```gherkin
When the user attempts to delete the `latest` PromiseRevision
Then the deletion is blocked with a warning that that PromiseRevision cannot be deleted
```

ChunyiLyu:
# Context

closes #638

In action:
```
❯ k get promiserevisions
NAME           LATEST
redis-v0.1.0
redis-v0.2.0   true
❯ k delete promiserevisions.platform.kratix.io redis-v0.1.0
promiserevision.platform.kratix.io "redis-v0.1.0" deleted
❯ k delete promiserevisions.platform.kratix.io redis-v0.2.0
Error from server (Forbidden): admission webhook "vpromiserevision-v1alpha1.kb.io" denied the request: can not delete the latest PromiseRevision
```

Delete validation will only block latest promise revision deletion if the action is *not* performed by kratix controller (by checking if the service account is from the kratix platform system namespace) or by kubernetes generic garbage collector. This means that promise controller can continue cleaning up revisions at promise deletion.

### DIFF
--- a/internal/webhook/v1alpha1/promiserevision_webhook.go
+++ b/internal/webhook/v1alpha1/promiserevision_webhook.go
@@ -1,19 +1,3 @@
-/*
-Copyright 2021 Syntasso.
-
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
-
-    http://www.apache.org/licenses/LICENSE-2.0
-
-Unless required by applicable law or agreed to in writing, software
-distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


jeduardo (changes requested):
Just a couple of nitpicks

PR #648, opened 2025-12-15
chore: add index for ResourceBinding lookup

### LINKED ISSUE #647: chore: improve lookup for ResourceBindings when deleting ResourceRequests for a PromiseRevision
# Context

https://github.com/syntasso/kratix/pull/645 highlighted that our implementation currently looks for all ResourceBindings for a Promise and selects the relevant ones based on the PromiseRevision version.

We should look into a way to make Kubernetes do the work of selecting the relevant ResourceBindings for us based on the promise-name label and the desired version.

jeduardo:
This change:
- adds an index over the `spec.version` field for ResourceBindings
- uses the index to retrieve all relevant ResourceBindings for a PromiseRevision based on promise name and PromiseRevision version
- removes a version-specific check as the code will now operate only over relevant bindings
- adds index support to our test k8s client

Closes #647

### DIFF
--- a/internal/controller/suite_test.go
+++ b/internal/controller/suite_test.go
@@ -81,18 +81,29 @@ var _ = BeforeEach(func() {
 		return client.Status().Update(ctx, obj, opts...)
 	}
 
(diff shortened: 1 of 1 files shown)


SaphMB (approved):
Nice optimisation

PR #657, opened 2026-01-01
feat: add promise delete pipeline failed status condition

### LINKED ISSUE #412: feat: add promise delete pipeline failed status condition
### Summary

In https://github.com/syntasso/kratix/pull/362 we added a `DeleteWorkflowCompletedCondition` which is (1) currently only used on Resources, and (2) is only updated on Resource Delete workflow failures.

This issue is to address (1) by adding the same status condition to Promises when a Promise Delete workflow fails.

### Details

See https://github.com/syntasso/kratix/pull/362 for the Resource implementation; the Promise implementation is very similar!

majiayu000:
Fixes #412

## Changes
- Add DeleteWorkflowCompletedCondition to Promise status when delete pipeline fails
- Record a warning event on the Promise when delete pipeline fails
- Follow the same pattern as Resource implementation (PR #362)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -9,7 +9,7 @@ Before submitting code, please ensure your development environment is set up cor
 Before submitting your changes, please:
 
 1. [Install Go](https://go.dev/doc/install) and ensure the toolchain is available in your `$PATH`.
-2. Install [`golangci-lint`](https://golangci-lint.run/usage/install/).
+2. Install [`golangci-lint`](https://golangci-lint.run/docs/welcome/install/).
 3. Run `make test` to execute unit tests.
 4. Run `make lint` to run static analysis.
 5. Run `make system-test` to run the system tests.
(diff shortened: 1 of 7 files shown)


internal/controller/promise_controller.go
```diff
@@ -19,6 +19,7 @@ package controller
 
 import (
 	"context"
+	stderrors "errors"
```
shano: Not sure why including a standard error library is needed.

shano (changes requested):

internal/controller/promise_controller.go
majiayu000: The `k8s.io/apimachinery/pkg/api/errors` package is imported as `errors` in this file, which shadows the standard library's `errors` package. I explicitly aliased the standard library as `stderrors` to access `errors.Is` for checking `workflow.ErrDeletePipelineFailed`.

shano (approved):
LGTM

PR #659, opened 2026-01-01
feat: ability to set resource labels from a workflow

### LINKED ISSUE #617: feat: ability to set resource labels from a workflow
**As a** platform owner (and maybe promise author)
**I want** to update my resource with some labels or annotations
**So that** I can integrate with my systems

## Context

We allow resource status to be updated from the pipeline, but provide no mechanism for labels or annotations. Allowing those to be set is useful for our users.

## Acceptance
```gherkin
Scenario: labels from the workflow
Given a stage that uses the labelling feature
When I apply the resource request and the pipeline completes
Then the resource includes the dynamically generated labels

Scenario: GO SDK
Given I'm building stages with the Go sdk
When I want to apply labels to my resource
Then I can use a helper function

Scenario: Python SDK
Given I'm building stages with the Python sdk
When I want to apply labels to my resource
Then I can use a helper function
```

## Tasks

- [ ] Update docs

majiayu000:
Fixes #617

## Changes
- Add MergeLabels function to merge incoming labels with existing resource labels
- Modify update-status command to read and apply labels from /kratix/metadata/labels.yaml
- Update SDK contract with WriteLabels and ReadLabels function specifications
- Add tests for label merging functionality

### DIFF
--- a/sdk/contract.md
+++ b/sdk/contract.md
@@ -26,6 +26,10 @@ WriteOutput writes the content to the specifies file at the path /kratix/output/
 
 WriteStatus writes the specified status to the `/kratix/metadata/status.yaml`
 
+**`WriteLabels(map[string]string) error`**
+
+WriteLabels writes the specified labels to the `/kratix/metadata/labels.yaml`. These labels will be merged with the existing resource labels after the pipeline completes.
+
 **`WriteDestinationSelectors([]DestinationSelector) error`**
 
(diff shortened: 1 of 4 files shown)


shano:
Thank you for the contribution, but this doesn't reflect the design direction we're looking for, by adding new files between pipeline stages. 

We're not looking for improvements in this area at the moment, so I'll remove the good-first-issue label as I think this needs some further discussions internally on how to deliver this and close your PR, but thanks again for the submission!

Issue #628, opened 2025-10-27
bug: healthrecord fails to delete after promise is deleted


kirederik:
## Steps to reproduce

1. Create a Promise
2. Create a resource request
3. Create a health record for the resource request
4. Delete the promise
5. Wait for promise deletion to finish
6. delete health records and observe that:
* health record is stuck at deletion
* there is no error in operator logs, only reconciliation finished logs:
```
{"level":"info","ts":"2026-01-13T13:48:18Z","logger":"controllers.HealthRecordController","msg":"reconciliation started","controller":"healthRecord","name":"healthy-example","namespace":"default","promise":"healthchecktest","resourceName":"example","resourceNamespace":"default","generation":2,"severity":"info"}
{"level":"info","ts":"2026-01-13T13:48:18Z","logger":"controllers.HealthRecordController","msg":"reconciliation finished","controller":"healthRecord","name":"healthy-example","namespace":"default","promise":"healthchecktest","resourceName":"example","resourceNamespace":"default","generation":2,"severity":"info","duration":0.000030042}
```

## Expected

1. Health Record is deleted (it will eventually be removed from the state store by the health agent in the worker)

## Actual

1. Health Record hangs at deletion

## Why
The reason that the health records hangs at deletion is this block: https://github.com/syntasso/kratix/blob/main/internal/controller/healthrecord_controller.go#L76-L80

Health records controller tries to get the promise, in our case the promise is gone, so the controller receives a 'not found' error, which is then ignored at the return value. We return the reconciliation with no error, but health records finalizer is still there and therefore the deletion hangs.

PR #668, opened 2026-01-08
fix: ensure health records are deleted before promise dependencies

majiayu000:
While fixing the "stuck finalizer" issue in PR #658, we identified that the root cause was the deletion order.

This PR enforces that all HealthRecords associated with a Promise are fully deleted (and their finalizers removed) BEFORE the Promise controller deletes the RBAC resources (ClusterRole/Bindings). This prevents HealthRecords from getting stuck in a "zombie" state where they cannot be finalized due to missing permissions.

Closes the architectural gap identified in #658 discussion.


ChunyiLyu:
Thanks for creating a PR @majiayu000. Is this one meant to fix issue #628? It that's the case, would you likely to close the previous PR you have associated with the same issue https://github.com/syntasso/kratix/pull/658?

PR #671, opened 2026-01-14
fix: switch from go-git to Git native client

### LINKED ISSUE #666: fix: use git cli for git operation
## Background

We have been experiencing intermittent Git errors when using Git as the Kratix state store (e.g. `packfile not found`, `object not found`, and other X not found errors). These errors have been difficult to reproduce consistently.

A spike was conducted to investigate the root cause and possible mitigations: https://github.com/syntasso/kratix-private/issues/168.

What we found in the spike is that using the go-git library, the errors can be reproduced reliably. When replacing go-git with the Git CLI for clone/commit/push operations eliminates the observed errors under the same workloads.

Based on these findings, we will replace the current git writer implementation based on go-git with an implementation that shells out to the Git CLI for Git operations.

## Goal
* Implement Git operations (clone, add, commit, push, permission validation) using the Git CLI.
* Support existing authentication modes:
   - SSH (private key + known_hosts)
   - Basic auth / GitHub App tokens (via HTTPS)
* Preserve existing GitStateStore behavior.
* There should be no change to the GitStateStore/Destinstion CRD schema. No user facing behavior change.

## Acceptance
* GitStateStore operations use the Git CLI instead of go-git.
* SSH key and basic auth auth works as before.
* Sensitive credentials are not logged or printed out at when failing.
* The previously observed `packfile not found` / `object not found` / `X not found` errors do not appear.

### Engineering notes
* argocd git client uses git cli: https://github.com/argoproj/argo-cd/blob/master/util/git/client.go#L166
* git cli uses other underlying library for SSH and TLS. We should ensure that user facing behavior stays consistent
* this solution should be properly soak tested

jeduardo:
This PR switches our usage of go-git for cloning and performing file management operations in Git repositories to use the native Git client instead. It:

- Imports a Git execution library from Argo CD to shell out to the native Git binary
- Rebases the build of the Kratix docker image to alpine/git, so all requirements are packaged with Kratix
- Adds an integration test suite for the Git client and Git writers

Closes #666


Dockerfile
```diff
@@ -20,19 +20,20 @@ RUN --mount=target=. \
     --mount=type=cache,target=/go/pkg/mod \
     go build -o /out/pipeline-adapter work-creator/*.go
 
-FROM busybox AS busybox
+#FROM busybox AS busybox
```
shano: Why not just removed these commented lines?

jeduardo:
> Is the plan to clean up the commit history after it's approved?

Definitely, I'll squash and merge history, summarising only the key changes in the extended description.

lib/writers/git.go
```diff
@@ -35,184 +20,68 @@ type GitWriter struct {
...
+		git.GitClientRequest{
+			RawRepoURL: stateStoreSpec.URL,
+			Root:       repoPath,
+			Auth:       auth,
+			// NOTE: intentionally allowing insecure connections,
+			// due to pre-existing setting:
+			// https://github.com/syntasso/kratix/blob/59231e70b0a4a428067e3b909fd2e9dc07110997/lib/writers/git.go#L373
+			Insecure: true,
+			Proxy:    "",
+			NoProxy:  "",
```
kirederik:
```suggestion
```

lib/writers/git.go
```diff
@@ -316,87 +190,51 @@ func (g *GitWriter) deleteExistingFiles(removeDirectory bool, dir string, worklo
...
 		return nil, err
 	}
 	return content, nil
 }
 
-func (g *GitWriter) setupLocalDirectoryWithRepo(logger logr.Logger) (string, *git.Repository, *git.Worktree, error) {
-	localTmpDir, err := createLocalDirectory(logger)
+// Initialise a local directory with the Git repository and returns
+// localDir that should be cleaned up after use.
+func (g *GitWriter) setupLocalDirectoryWithRepo() (string, error) {
```
kirederik: nit: there's very little benefit to this function now

test/git/git_test.go
```diff
@@ -0,0 +1,1068 @@
...
+					By("creating a new file", func() {
+						file, err := os.Create(path)
+						Expect(err).ToNot(HaveOccurred())
+						// Defensive: if the file exists as a leftover from a previous test, the commit will still go
+						randomContent := fmt.Sprintf("random-%d\n", rand.Int())
+						_, err = file.WriteString(randomContent)
+						Expect(err).ToNot(HaveOccurred())
+						err = file.Close()
+						Expect(err).ToNot(HaveOccurred())
+					})
```
kirederik: nit: dupication; can we add this to a helper function?

test/git/git_test.go
```diff
@@ -0,0 +1,1068 @@
...
+						Expect(err.Error()).To(ContainSubstring("src refspec another-branch does not match any"))
+					})
+
+					By("cleaning out the test file", func() {
+						err = os.Remove(path)
+						Expect(err).ToNot(HaveOccurred())
+					})
+				})
+
+				When("multiple clients are pushing to the same repo", func() {
```
kirederik: question: why is this test relevant to our system?

kirederik (changes requested):
see comments!

test/git/git_test.go
jeduardo: We added this to cover the scenario where multiple destinations are using the same GitStateStore but under different paths. In this case, we will have works being saved as different files under different directories (without conflict) but the HEAD of the branch will change. go-git handled this by automatically running a pull before attempting a commit and push, safely remediating the situation.

The Git binary does not does this by default and, instead, issues a warning informing that the branch isn't fresh. In our implementation, we replicated go-git's behaviour and triggered a pull before doing a commit and running a push.

As we implemented this behaviour by ourselves, we wanted to make sure it was described and tested.

PR #716, opened 2026-03-02
feat: Log clearly what resource request a reconciliation belongs to

aclevername:
this is only for work/workplacement/dynamic/resource-binding controllers. Introduces a new `<promise-name>/<resource-namespace>/<resource-name>` log key, that makes filtering easier, particularly across controllers


aclevername:
> But it does look _slightly_ strange and potentially confusing when promises are being reconciled as the "resourceRequest" is detailed as promiseName/Namespace/.
> 
> ```
> "promise": "redis", "resourceRequest": "redis/kratix-platform-system/", "severity": "info", "duration": "294.5µs"}
> ```
> 
> The key being absent or nil could alleviate this

Good point! Updating so that the work/workplacement controller omit this label when it isn't associated to a resource request

PR #724, opened 2026-03-05
chore: clearer logging msgs in promise and dynamic controller

ChunyiLyu:


internal/controller/promise_controller.go
```diff
@@ -214,9 +214,10 @@ func (r *PromiseReconciler) Reconcile(ctx context.Context, req ctrl.Request) (re
 			return result, statusUpdateErr
 		}
 		if originalStatus == v1alpha1.PromiseStatusAvailable {
+			msg := "Promise no longer available: Requirements have changed"
```
senglezou: nice, thanks for this, makes it clearer

Issue #736, opened 2026-03-17
bug: promise revisions cannot create successfully when promise version contains certain characters


ChunyiLyu:
## Summary

We the promise name and the promise version to construct promise revisions names. Promise versions are labels. Labels allow a different set of characters compare to names, which lead to the case where you can install the promise successfully but its promise revisions fail to be created.

```
{"level":"error","ts":"2026-03-17T16:37:08Z","msg":"Reconciler error","controller":"promise","controllerGroup":"platform.kratix.io","controllerKind":"Promise","Promise":{"name":"teststatus"},"namespace":"","name":"teststatus","reconcileID":"c8ac1aca-48ad-41a2-b0c5-6da8dbaf99ed","error":"PromiseRevision.platform.kratix.io \"teststatus-A-GREAT-VERSION\" is invalid: metadata.name: Invalid value: \"teststatus-A-GREAT-VERSION\": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')","stacktrace":"sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).reconcileHandler\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.20.4/pkg/internal/controller/controller.go:347\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).processNextWorkItem\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.20.4/pkg/internal/controller/controller.go:294\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller[...]).Start.func2.2\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.20.4/pkg/internal/controller/controller.go:255"}
```

### To reproduce

1. install an example promise with upper case in its version:
```
apiVersion: platform.kratix.io/v1alpha1
kind: Promise
metadata:
  name: teststatus
  labels:
    kratix.io/promise-version: A-GREAT-VERSION
spec:
  api:
    apiVersion: apiextensions.k8s.io/v1
    kind: CustomResourceDefinition
    metadata:
      name: teststatuses.test.kratix.io
    spec:
      group: test.kratix.io
      names:
[...]

PR #748, opened 2026-03-21
fix(#736): ensure promise revision name is valid

kirederik:
We the promise name and the promise version to construct promise
revisions names. Promise versions are labels. Labels allow a different
set of characters compare to names, which lead to the case where you can
install the promise successfully but its promise revisions fail to be
created.

This commit removes the version label from the revision name entirely.
Instead, it generates a hash based on the version, and append that to
the promise name.

To maintain readability and usability, this commit also adds extra
information to the promise revision printer columns

solves #736

### DIFF
--- a/internal/controller/promise_controller_test.go
+++ b/internal/controller/promise_controller_test.go
@@ -1978,7 +1978,16 @@ var _ = Describe("PromiseController", func() {
 						}),
 					})).To(Succeed())
 					Expect(revisionList.Items).To(HaveLen(2))
-					revision := revisionList.Items[1]
+					var revision v1alpha1.PromiseRevision
+					found := false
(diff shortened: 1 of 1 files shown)


kirederik:
> Implementation looks good but I'm wondering what we want to do for revisions created with before the upgrade? I see them being left around with the same promise version label:
> 
> ```
> ❯ k get promiserevisions --show-labels
> NAME           PROMISE   VERSION   LATEST   LABELS
> redis-77ba1    redis     v0.1.0    true     kratix.io/latest-revision=true,kratix.io/promise-name=redis
> redis-v0.1.0   redis     v0.1.0             kratix.io/promise-name=redis
> ```
> 
> What do you think?

Good catch! we updated the code to only create a revision when another one is not present.

PR #754, opened 2026-03-26
Allow labels cascading from Promise object into dynamic ClusterRoles

antonioquintavalle1A:
Hello folks,
with this commit I'd like to request a change in the way how Kratix manages dynamic ClusterRoles created from brand new Promises.
The logic here is to cascade any labels contained into the promise.spec.clusterrolelabels directly into the clusterrole.spec.metadata.labels.

Main benefit of this approach is to benefit from the dynamic RBAC feature in kubernetes where we aggregate new capabilities to existing roles / rolebindings.

If no labels are specified, the behavior is kept the same.

Please do not hesitate to share your thoughts on this.


kirederik (commented):
Thanks for your contribution @antonioquintavalle1A !

I left a few comments in the PR. Would you be able to address those? Also you would need to sign the CLA for us to get it merged.

I'm also curious if you have considered other ways to define the labels that are not promise based? I'm wondering if this should be a platform level configuration vs a promise level one...

kirederik (changes requested):

kirederik:
Thanks for the contribution @antonioquintavalle1A! As we discussed, we intend to make this a more generic change you can apply to the resources. Stay tuned!

PR #782, opened 2026-05-13
fix: improve controller concurrency

Phiph:
This pull request introduces several improvements aimed at enhancing the performance and scalability of the Kratix controller. The changes include new configuration knobs for controller concurrency and Kubernetes API rate limiting, documentation of performance findings, and a new Makefile target for running performance tests. 

**Performance and Scalability Improvements**

- Added new command-line flags in `cmd/main.go` to control `MaxConcurrentReconciles` for dynamic resource-request controllers, filter out no-op status update events, and configure Kubernetes API QPS/burst limits. These allow operators to tune controller concurrency and API usage for large clusters. [[1]](diffhunk://#diff-c444f711e9191b53952edb65bfd8c644419fc7695c62611dc0fb304b4fb197d6R120-R144) [[2]](diffhunk://#diff-c444f711e9191b53952edb65bfd8c644419fc7695c62611dc0fb304b4fb197d6R170-R185) [[3]](diffhunk://#diff-c444f711e9191b53952edb65bfd8c644419fc7695c62611dc0fb304b4fb197d6R200) [[4]](diffhunk://#diff-c444f711e9191b53952edb65bfd8c644419fc7695c62611dc0fb304b4fb197d6R250) [[5]](diffhunk://#diff-c444f711e9191b53952edb65bfd8c644419fc7695c62611dc0fb304b4fb197d6R326-R327)

- Implemented an `applyKubeAPIRateLimits` function to set client-go QPS and burst values based on the new flags, helping to prevent client-side throttling under high write loads. [[1]](diffhunk://#diff-c444f711e9191b53952edb65bfd8c644419fc7695c62611dc0fb304b4fb197d6R120-R144) [[2]](diffhunk://#diff-c444f711e9191b53952edb65bfd8c644419fc7695c62611dc0fb304b4fb197d6R200) [[3]](diffhunk://#diff-c444f711e9191b53952edb65bfd8c644419fc7695c62611dc0fb304b4fb197d6R250)

**Testing and Benchmarking**

- Introduced a `perf-test` Makefile target to run the controller performance rig with customizable parameters (number of requests, run name, context, etc.), making it easier to benchmark controller performance in different scenarios.

# Perf flag comparison — N=100 RRs, no-op Promise

8 runs, single-cluster kind, fresh install. Measures the effect of three
controller knobs in isolation and combination:

- `--dynamic-rr-max-concurrent-reconciles` (MCR): 0 (=1) vs 20
- `--kube-api-qps` / `--kube-api-burst`: client-go defaults (20 / 30) vs 200 / 400
- `--dynamic-rr-filter-noop-writes` (predicate): off vs on

All runs target the same no-op Promise — one `busybox sh -c "exit 0"` configure
pipeline that produces zero manifests. Convergence is wall-clock from the last
RR `Create` to every RR carrying `Reconciled=True`.

## Headline numbers

| run | MCR | QPS / Burst | predicate | wall | p50 recon | longest_running peak (s) | unfinished_work peak (s) | rec total CPU (s) |
|---|---:|---|---|---:|---:|---:|---:|---:|
| mcr1-flagsoff | 1 | 20 / 30 | off | **2m59s** | 11 | 1.12 | — | 178.7 |
| mcr1-qps | 1 | 200 / 400 | off | **2m3s** | 12 | 1.13 | — | 122.6 |
| mcr1-predicate | 1 | 20 / 30 | on | **2m59s** | 11 | 1.16 | — | 178.7 |
| mcr1-all | 1 | 200 / 400 | on | **2m5s** | 12 | 1.12 | — | 123.7 |
| mcr20-flagsoff | 20 | 20 / 30 | off | **1m40s** | 14 | 3.13 | 26.7 | 875.5 |
| mcr20-qps | 20 | 200 / 400 | off | **38s** | 18 | 1.07 | 6.8 | 69.5 |
| mcr20-predicate | 20 | 20 / 30 | on | **1m38s** | 14 | 2.97 | 27.0 | 778.2 |
| mcr20-all | 20 | 200 / 400 | on | **40s** | 18 | 0.40 | 3.3 | **55.4** |

(`unfinished_work_seconds` is the sum of seconds the perftest workqueue has had
items in-flight — a proxy for "how long are workers stalled?". Empty for MCR=1
runs because at concurrency=1 the metric never accumulates.)

### MCR=20 alone

1m40s vs 2m59s baseline (1.78× faster wall-clock). But each reconcile got
**~3× slower** (`longest_running` 1.12s → 3.13s) and total controller CPU
spent reconciling went from 178s to **875s** — almost 5× more work, for less
than 2× the throughput.

Cause: parallel goroutines compete for the same 20-QPS rate-limited kube-API
client. Each Status().Update stalls in the token bucket, multiplied across
20 concurrent workers. `unfinished_work_seconds` peaks at **26.7s** — workers
spend an aggregate ~27 seconds stuck mid-reconcile waiting for API budget.

**Verdict**: MCR by itself is throughput-positive but CPU-negative. Don't
ship it without raising QPS.

### QPS=200/400 alone (MCR=1)

2m3s vs 2m59s (1.45× faster) at MCR=1 with the same reconcile count.
At MCR=1 the controller is a serial loop; raising QPS just makes each
write faster. CPU total drops from 178s to 123s — pure efficiency.

But: 100 retries appeared in `workqueue_retries_total` (was 0 in baseline).
Each retry is a queue add with backoff. At MCR=1 baseline, status writes
re-enqueue via the self-watch deterministically; at higher QPS, the
controller occasionally hits something (likely a conflict on Status update)
that triggers a retry. **Worth a closer look** — the retries didn't hurt
wall-clock at this N but could compound at scale. Same retry count appears
in `mcr1-all`, so this is QPS-coupled, not predicate-coupled.

### Predicate alone

**No measurable wall-clock difference in either MCR setting.** At MCR=1 the
state machine is gated by serial reconciles; the predicate has nothing to
filter (every status write is genuinely advancing state, no spurious
re-fires to drop). At MCR=20 the bottleneck is API-rate, not event volume.

### Predicate's hidden value (look at the metrics)

The wall-clock table understates the predicate. Compare MCR=20 pairs:

|  | adds (queue events) | reconcile_count | rec total CPU (s) |
|---|---:|---:|---:|
| mcr20-flagsoff | 566 | 456 | 875.5 |
| mcr20-predicate | **518 (−48)** | **409 (−47)** | **778.2 (−97s)** |
| mcr20-qps | 520 | 400 | 69.5 |
| mcr20-all | **491 (−29)** | **399 (−1)** | **55.4 (−14s)** |

Two observations:

1. **The predicate is filtering events.** ~50 queue adds and ~50 reconciles
   get suppressed per 100 RRs at MCR=20 (predicate vs flagsoff). At MCR=20+QPS
   it's smaller (~30) because the controller burns through state-machine
   steps faster, so fewer redundant self-watch events have time to accumulate.
2. **The CPU win is disproportionate.** A 10% drop in reconcile count gives
   an 11% drop in CPU at MCR=20+throttled, and **a 20% drop at MCR=20+QPS**.
   That's because the *throttled* runs spend most of their reconcile time
   stalled in the rate-limiter (not in actual work), so cutting 10% of
   reconciles only saves 10% of CPU. The unthrottled runs do all their CPU
   work for real, so the 10% you cut is pure savings.

The predicate is **only worth pairing with the QPS bump**.
On its own, it cuts a fraction of stalls that the rate-limiter would have
absorbed anyway. With QPS unblocked, every filtered reconcile is real CPU
saved.

The `longest_running` metric on mcr20-all (0.40s) is the cleanest measure of
how fast the controller is in its best state: reconciles complete in 400ms
and there are no parallel-worker stalls.


Phiph:
I found that Flux's default is 50/300, and I wanted to test something in between two other extremes (MCR=1 baseline and MCR=20 flat-out).

Flags used:

```
--dynamic-rr-max-concurrent-reconciles=10
--kube-api-qps=50
--kube-api-burst=300
--dynamic-rr-filter-noop-writes=true
```

#### Result

| run | MCR | QPS / Burst | predicate | wall-clock | p50 recon | longest_running peak |
|---|---:|---|---|---:|---:|---:|
| mcr1-flagsoff | 1 | 20 / 30 | off | 2m59s | 11 | 1.12s |
| mcr1-all | 1 | 200 / 400 | on | 2m5s | 12 | 1.12s |
| mcr20-flagsoff | 20 | 20 / 30 | off | 1m40s | 14 | 3.13s |
| mcr20-all | 20 | 200 / 400 | on | **40s** | 18 | 0.40s |
| **mcr10-qps50** *(new)* | **10** | **50 / 300** | **on** | **52s** | **16** | **1.29s** |

- A halved budget gave most of the performance on this run — MCR 20→10 and QPS 200→50 only added ~12s vs the 40s run.
- `longest_running` came down to ~1.3s, similar to the MCR=1 numbers, so workers don't look like they're queued behind the rate-limiter at QPS=50.
- Same reconciles-per-RR climb with parallelism we've seen before (p50: 11 → 16 → 18 across MCR=1/10/20). Predicate is on here too — it trims a little but doesn't change the shape.

Not pushing for a specific default — just adding the data point. Happy to run other points on the curve if anyone has a config in mind.

internal/controller/dynamic_resource_request_controller.go
```diff
@@ -282,8 +282,22 @@ func (r *DynamicResourceRequestController) Reconcile(ctx context.Context, req ct
...
+	// persisted, so we can coalesce them into a single Status().Update at the
+	// end of Reconcile instead of writing once per mutation. Each write fires
+	// the RR self-watch and re-enqueues this reconcile, so coalescing
+	// directly reduces reconciles-per-RR.
+	//
+	// Note: B17 (observedGeneration) is *not* coalesced into this — its
+	// write-and-return is part of the controller's external contract (clients
+	// rely on the reconcile that bumps observedGeneration returning ctrl.Result{}
+	// rather than the longer-tail nextReconciliation result). Changing it
+	// breaks `dynamic_resource_request_controller_test.go:406`.
```
kirederik: what ire b17 & b19? 👀 not sure this comments are very relevant (also it references a line of code, it will get stale fast)

internal/controller/predicate_test.go
```diff
@@ -0,0 +1,158 @@
+//nolint:testpackage // unit-tests the unexported dynamicRRNoOpWriteFilter
+package controller
```
kirederik: can this be ginkgo?

kirederik (changes requested):
Thanks for your contribution Phill, and sorry for the delay. Can you take a look at the comments above?

Also, the `perf` test suite is massive 😅 I'll spend a bit more time going through it and running a few times!

internal/controller/predicate_test.go
Phiph: I'll take a look.

internal/controller/dynamic_resource_request_controller.go
Phiph: I asked AI to generate me some notes on the reconciliation loop to analyse it and refer to it as I built. I'll remove my notes.

PR #790, opened 2026-05-18
fix: ensure UpgradeFailed propagates to resource bindings

kirederik:


senglezou (approved):
I rerun the failing test so you can merge once it goes green. It's just a flake

senglezou:
I know I approved, but won't reject the changes as not to block the PR. But after spending time on looking why the test is failing, this new logic seems to not be quiet right.

PR #801, opened 2026-05-26
feat(789): bind resources to version

### LINKED ISSUE #789: feat: support the ability to bind all resources to a specific version
## Context
https://miro.com/app/board/uXjVHQYYTMk=/?share_link_id=51056481832 
Once people turn on promise upgrade feature flag, promise revisions and resource bindings are created automatically by users. By default, resource bindings are all created to bind to the 'latest' version of promises (`spec.version` is set to `latests`).

There's feedback that we should allow a platform-wide setting where resource bindings are created to bind to a specific version, instead of setting to `latest`.

### Eng notes

We generally support platform wide setting in kratix configmap. This will also be something we enable from there.
Name option if you can't think of anything better
```yaml
resourceBindings:
  defaultVersion: [floating | pinned] # floating is the current and default behavior
```

Updates between different strategy will only impact new bindings. Kratix will not update existing resource bindings when the default version strategy changes.

Note we are dealing with the platform-level only here. We will need to find a mechanism to support other methods in the future, see miro board.

## Acceptance
```gherkin
Scenario: Configuration set to floating
Given I have defaultVersion set to floating
When I request a new resource on my platform
Then the resource binding's spec.version is set to "latest"

Scenario: Configuration set to pinned
Given I have defaultVersion set to pinned
When I request a new resource on my platform
Then the resource binding's spec.version is set to the spec.version in the PromiseRevision

Scenario: Configuration not set
Given I have have not set the defaultVersion configuration
When I request a new resource on my platform
Then the resource binding's spec.version is set to "latest"
```

SaphMB:
closes #789 

Example configmap:

```
apiVersion: v1
data:
  config: |
    logging:
      level: "trace" # one of info, warning, debug, trace
      structured: false # if true, emit logs as jso
    featureFlags:
      promiseUpgrade: true # enable/disable promise revisions
    ResourceBindingVersionStrategy: "pinned"
kind: ConfigMap
metadata:
  name: kratix
```

- **feat: ResourceBindings.DefaultVersion can be defined via the kratix configmap**
- **feat: set the resourceBinding.spec.version in accordance with the defaultVersion**
- **feat: provide feedback when an invalid defaultversion is provided**
- **feat: log the defaultVersion in startup**
- **chore: define system test for defaultVersion behaviour**

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -73,19 +73,15 @@ func init() {
 }
 
 type KratixConfig struct {
-	Workflows                Workflows               `json:"workflows"`
-	NumberOfJobsToKeep       int                     `json:"numberOfJobsToKeep,omitempty"`
-	ControllerLeaderElection *LeaderElectionConfig   `json:"controllerLeaderElection,omitempty"`
-	SelectiveCache           bool                    `json:"selectiveCache,omitempty"`
-	ReconciliationInterval   *metav1.Duration        `json:"reconciliationInterval,omitempty"`
-	Telemetry                *telemetry.Config       `json:"telemetry,omitempty"`
-	Logging                  *LoggingConfig          `json:"logging,omitempty"`
-	FeatureFlags             *FeatureFlags           `json:"featureFlags,omitempty"`
-	ResourceBindings         *ResourceBindingsConfig `json:"resourceBindings,omitempty"`
-}
-
-type ResourceBindingsConfig struct {
-	DefaultVersion ResourceBindingDefaultVersion `json:"defaultVersion,omitempty"`
+	Workflows                      Workflows                     `json:"workflows"`
+	NumberOfJobsToKeep             int                           `json:"numberOfJobsToKeep,omitempty"`
+	ControllerLeaderElection       *LeaderElectionConfig         `json:"controllerLeaderElection,omitempty"`
+	SelectiveCache                 bool                          `json:"selectiveCache,omitempty"`
+	ReconciliationInterval         *metav1.Duration              `json:"reconciliationInterval,omitempty"`
+	Telemetry                      *telemetry.Config             `json:"telemetry,omitempty"`
+	Logging                        *LoggingConfig                `json:"logging,omitempty"`
+	FeatureFlags                   *FeatureFlags                 `json:"featureFlags,omitempty"`
+	ResourceBindingVersionStrategy ResourceBindingDefaultVersion `json:"resourceBindingVersionStrategy,omitempty"`
 }
 
 // ResourceBindingDefaultVersion controls the version strategy for ResourceBindings.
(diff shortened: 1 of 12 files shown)


internal/controller/dynamic_resource_request_controller.go
```diff
@@ -584,12 +585,19 @@ func (r *DynamicResourceRequestController) updateResourceBinding(ctx context.Con
...
+			} else if r.ResourceBindingPinned {
+				revision, err := latestRevision(ctx, r.Client, promise)
+				if err != nil {
+					return err
+				}
+				resourceBinding.Spec.Version = revision.Spec.Version
+			} else {
+				resourceBinding.Spec.Version = LatestVersion
 			}
 		}
```
aclevername, line 595: worth a quick helper functon? make the code bit easier to read, if/elseif/else is a mouthful

cmd/main_test.go
```diff
@@ -0,0 +1,77 @@
+/*
```
aclevername, line 1: migrate to ginkgo

cmd/main.go
```diff
@@ -283,6 +300,7 @@ func main() {
 		ReconciliationInterval: getRegularReconciliationInterval(kratixConfig),
 		EventRecorder:          mgr.GetEventRecorderFor("PromiseController"),
 		PromiseUpgrade:         promiseUpgradeEnabled(kratixConfig),
+		ResourceBindingPinned:  resourceBindingDefaultVersion == ResourceBindingDefaultVersionPinned,
```
aclevername, line 299: this nicely points out that maybe the value should be a bool? or do we envisage more possible values?

internal/controller/dynamic_resource_request_controller_test.go
```diff
@@ -1332,6 +1359,24 @@ var _ = Describe("DynamicResourceRequestController", func() {
...
+						Expect(fakeK8sClient.Status().Update(ctx, resReq)).To(Succeed())
+
+						result, err := t.reconcileUntilCompletion(reconciler, resReq)
+						Expect(err).NotTo(HaveOccurred())
+						Expect(result).To(Equal(ctrl.Result{}))
+
+						binding := getResourceBinding(promise.GetName(), resReqNameNamespace)
+						Expect(binding.Spec.Version).To(Equal(promiseVersion))
+					})
+				})
```
aclevername, line 1391: i think this makes sense, but lets chat to double check

test/system/resource_binding_default_version_test.go
```diff
@@ -0,0 +1,84 @@
+package system_test
```
aclevername, line 1: does this change warrant a dedicated sys test suite?

cmd/main.go
SaphMB: There's a possibility that there may be options so I'm reluctant to assume just two

aclevername:
@codex review

PR #806, opened 2026-05-28
feat: resource binding can be reconciled w/ label

kirederik:
closes [#1006](https://github.com/syntasso/enterprise-kratix/issues/1006)


internal/controller/resourcebinding_controller.go
```diff
@@ -118,49 +120,71 @@ func (r *ResourceBindingReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+		}
 		apiMeta.SetStatusCondition(&resourceBinding.Status.Conditions, metav1.Condition{
 			Type:               v1alpha1.UpgradeSucceededCondition,
 			Status:             metav1.ConditionUnknown,
 			Reason:             v1alpha1.UpgradeInProgressReason,
-			Message:            fmt.Sprintf("Upgrade to version %s is in progress", desiredVersion),
+			Message:            statusMessage,
 			LastTransitionTime: metav1.Now(),
 		})
 		resourceBinding.Status.FailedVersion = ""
```
senglezou, line 182: should we clear the Status.FailedVersion first before we delete the ManualReconciliationLabel on line 164? I can imagine we delete the label and by the time we get here to clean up the status if something happens and we don't clear it up it will stay there no?

internal/controller/resourcebinding_controller.go
```diff
@@ -118,49 +120,71 @@ func (r *ResourceBindingReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+		if upgradeSucceededCond != nil &&
+			upgradeSucceededCond.Status == metav1.ConditionFalse &&
+			resourceBinding.Status.FailedVersion == desiredVersion {
+			return ctrl.Result{}, nil
+		}
+		needsStatusUpdate = upgradeSucceededCond == nil ||
+			upgradeSucceededCond.Status != metav1.ConditionUnknown ||
+			upgradeSucceededCond.Reason != v1alpha1.UpgradeInProgressReason ||
+			resourceBinding.Status.FailedVersion != ""
+
```
SaphMB:
```suggestion
```

internal/controller/resourcebinding_controller.go
kirederik: I thought about it, but then we have the opposite problem: if updating the status fails, the manual reconciliation label is never removed, which will cause us to trigger the pipelines again; imo the manual reconciliation label should be deleted as soon as the resource labelling succeeds. 🤔

senglezou (approved):

PR #808, opened 2026-06-01
fix: Ensure status is updated when removing the reconcilitation labels

aclevername:
paired with @senglezou 

Solves two issues:
1. If a workflow is inflight, and its currently for the version we want to be upgrading to, dont apply the manual reconcilation label. Issue to improve it in the longterm https://github.com/syntasso/kratix/issues/814
1. After applying a manaul reconcilation label, ensure the status is updated. Previously the status update would fail since we just updated the binding to not have the label, the version of the binding we have is stale, so status update fails. Ideally we wouldn't fetch bt instead requeue, but we have no way to know if we've already applied a label, so this would trigger an infinite requeue and label scenario.

### DIFF
--- a/api/v1alpha1/resourcebinding_types.go
+++ b/api/v1alpha1/resourcebinding_types.go
@@ -17,6 +17,10 @@ limitations under the License.
 package v1alpha1
 
 import (
+	"strings"
+	"unicode"
+
+	apiMeta "k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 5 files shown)


api/v1alpha1/resourcebinding_types.go
```diff
@@ -94,3 +98,33 @@ type ResourceBindingList struct {
...
+// InFlightVersion returns the promise version of the resource configure
+// workflow that is currently being attempted for this binding, or an empty
+// string if no upgrade is in flight.
+//
+// It is inferred from the UpgradeSucceeded condition: while an upgrade is in
+// progress that condition is set to Unknown with reason UpgradeInProgress, and
+// its message embeds the desired version after the literal token "version ".
+// The controller is the sole writer of that condition, so coupling the parser
+// to the message format is acceptable as long as both stay in step.
+func (rb *ResourceBinding) InFlightVersion() string {
```
aclevername, line 111: this is awful, but its the cheaper approach to unblock the story currently. we can debate whether we are happy with this or not

internal/controller/resourcebinding_controller.go
```diff
@@ -139,37 +138,47 @@ func (r *ResourceBindingReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+	if manualReconciliationRequested {
 		delete(resourceBinding.Labels, resourceutil.ManualReconciliationLabel)
 		if err := r.Client.Update(ctx, resourceBinding); err != nil {
 			return ctrl.Result{}, err
 		}
+
+		//refetch so that we can update the status successfully afterwards without hitting a resource version conflict
+		if err := r.Client.Get(ctx, req.NamespacedName, resourceBinding); err != nil {
+			return ctrl.Result{}, err
+		}
```
aclevername, line 176: this fixes a bug where the status never gets updated after the manaul reconciliation label is applied. its not ideal, but we dont have an alternative currently

aclevername:
@codex review

senglezou (approved):

PR #834, opened 2026-07-02
feat: Add ability to wait in delete workflows

abangser:
This mirrors behaviour in the configure workflows. 

docs in https://github.com/syntasso/kratix-docs/pull/559

### DIFF
--- a/lib/resourceutil/util.go
+++ b/lib/resourceutil/util.go
@@ -23,7 +23,7 @@ const (
 	DeleteWorkflowCompletedCondition       = clusterv1.ConditionType("DeleteWorkflowCompleted")
 	DeleteWorkflowCompletedFailedReason    = "DeleteWorkflowFailed"
 	DeleteWorkflowSuspendedReason          = "DeleteWorkflowSuspended"
-	DeleteWorkflowSuspendedMessage         = "Delete pipeline is waiting before deleting Works"
+	DeleteWorkflowSuspendedMessage         = "The delete workflow is suspended"
 	PipelinesExecutedSuccessfully          = "PipelinesExecutedSuccessfully"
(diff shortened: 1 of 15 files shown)


aclevername:
@abangser shall we change this to a draft PR until its ready for review?

abangser:
Made sense @aclevername, had some niggles with the systems tests. But looking good now. Let me know if you (or anyone else) has any questions to talk through!

lib/resourceutil/util.go
```diff
@@ -22,6 +22,8 @@ const (
 	ConfigureWorkflowCompletedFailedReason = "ConfigureWorkflowFailed"
 	DeleteWorkflowCompletedCondition       = clusterv1.ConditionType("DeleteWorkflowCompleted")
 	DeleteWorkflowCompletedFailedReason    = "DeleteWorkflowFailed"
+	DeleteWorkflowSuspendedReason          = "DeleteWorkflowSuspended"
+	DeleteWorkflowSuspendedMessage         = "Delete pipeline is waiting before deleting Works"
```
ChunyiLyu: Not all configure workflow schedules work; maybe we change it to something a bit more generic?

ChunyiLyu (commented):
Looks good! My feedback is mostly on the e2e test. It think it currently covers resource workflow but not promise delete pipelines. Could we please add that? We could have one covers a simple suspend and the other one covers the retry case?

As this is a new feature and with side effects (delete jobs now will always have status writer), could you please also create docs PR?

lib/resourceutil/util.go
ChunyiLyu: updated

ChunyiLyu (approved):
:shipit:

PR #837, opened 2026-07-03
feat: warn when destination references missing statestore

### LINKED ISSUE #345: feat: destination webhook should give user a warning if it references a non-existing statestore
# Context

When user creates a Destination that references a non-existing statestore, users can find out when they check the destination status.
However, we would like a faster feedback than having to check its status. This issue is about introducing a destination webhook that gives users a warning if the referred State Store does not exist.

It's a warning and not an error because we don't want to create a hard ordering of people need to create State Store first.

Follow up from https://github.com/syntasso/kratix/pull/343

didiberman:
## Summary

Destinations can reference a `BucketStateStore` or `GitStateStore` that has not been created yet. Today, that only becomes visible later through Destination status, which makes the problem slower to diagnose.

This change adds an admission warning when a Destination references a missing StateStore. The warning gives immediate feedback during create/update, while still allowing the request so users are not forced to create resources in a strict order.

It also returns a clear validation error when `stateStoreRef` is omitted, instead of letting validation continue into code that expects the reference to exist.

Fixes #345.

## Testing

- `go run github.com/onsi/ginkgo/v2/ginkgo -r internal/webhook/v1alpha1`
- `make test`
- `PATH="$(go env GOPATH)/bin:$PATH" make lint`

### DIFF
--- a/internal/webhook/v1alpha1/destination_webhook.go
+++ b/internal/webhook/v1alpha1/destination_webhook.go
@@ -83,9 +83,14 @@ func (v *DestinationCustomValidator) ValidateCreate(ctx context.Context, destina
 		return nil, stderror.New("path field is required")
 	}
 
+	warnings, err := v.validateStateStoreRef(ctx, destination.Spec.StateStoreRef)
+	if err != nil {
+		return nil, err
+	}
+
 	// Check for path clashes with other destinations using the same state store
 	destinationList := &v1alpha1.DestinationList{}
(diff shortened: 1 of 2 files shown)


kirederik:
Thanks for your contribution @didiberman!

didiberman:
@kirederik I actually came across this issue while setting up my cluster with Kratix, MinIO and Backstage and so was great to contribute to the fix :)

PR #838, opened 2026-07-03
chore: remove act targets

### LINKED ISSUE #604: chore: remove act jobs from kratix project
The `Makefile` jobs `act-test-and-lint`, `act-integration-test`, and `act-system-test` are in a broken state since we externalised our jobs to `syntasso/ci/.github/workflows/test-kratix.yaml`. 

As workflows included with `uses` can be executed only as a full unit, there is no way to execute the tests independently as before. Running the workflow as single unit takes a long time and also cleans up local environments after runs. This removes most of the previous benefits of using `act` to execute individual CI jobs locally. 

We should remove the `act-*` tasks from our `Makefile` to avoid leaving unused and broken code in our builds.

didiberman:
## Summary

The local `act-*` Makefile targets no longer work now that CI jobs are externalized through the reusable `syntasso/ci` workflow. Keeping those targets advertises broken local commands and adds unused build logic.

This removes the stale `act` helper and the `act-test-and-lint`, `act-integration-test`, `act-system-test`, and `act-run-all` targets from the root Makefile.

Fixes #604.

## Testing

- `make help`
- `git diff --check`
- `make test`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -270,20 +270,6 @@ lint: golangci-lint # Lint with required config
 lint-all: # Lint with full config
 	golangci-lint run --config=.golangci.yml
 
-##@ act targets to run GH Action jobs locally
-
-ACT_JOB = act -j '$(1)' --rm -s GITHUB_TOKEN="$(shell gh auth token)"
-act-test-and-lint:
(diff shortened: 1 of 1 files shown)


kirederik:
Thank you!

PR #851, opened 2026-07-23
fix: recover from git rebase conflicts

### LINKED ISSUE #850: fix: GitStateStore writer cannot recover when the remote branch diverges
### Context

Kratix writes resources into a Git repository (the state store). If that repository is changed by something other than Kratix, e.g. a file deleted by a manual commit, Kratix can end up unable to write to that repository at all.

When it happens:

- Writes to that repository stop, for every resource using it, not just the one whose file changed.
- It doesn't recover on its own. Today the only fix is restarting the platform controller, which makes it start again from a clean copy of the repo.
- The affected `WorkPlacement` resources still report as healthy (`Ready`), so there's no obvious sign anything is wrong.

The cause is the same for both problems. Kratix keeps a local working copy of each repository and reuses it between writes, but it never refreshes that copy against the real remote before writing. Once the remote moves on, the local copy is stale: the write can't be reconciled with the remote, and the health status is judged against the stale local copy rather than what is really in the repo.

The impact is high: one file changed outside Kratix can block all updates to a destination, with a green status hiding it, until someone restarts the controller.

### Engineering notes

Background: the reusable local clone was added in PR #726 (`internal/controller/repository_cache.go`). The clone is only discarded when the state store's secret changes or the state store is deleted. Nothing calls `git fetch` on it between writes.

Two failures fall out of that:

1. **Stuck repository.** Before each write, `Reset()` (`lib/writers/git.go`) runs `git checkout --force <branch>` + `git clean -ffdx`. That clears uncommitted changes only. It does not move the local branch back to the remote, and it does not clear a half-finished rebase. The write commits locally and pushes; a rejected push falls back to `git pull --rebase` (`util/git/git.go`). If the remote deleted a file the local commit touches, the rebase hits a modify/delete conflict and stops, leaving a `.git/rebase-merge` directory that `Reset()` can't clean up. Every later write then fails, first with a rebase conflict and then with `already a rebase-merge directory`.

2. **False `Ready`.** The "did anything change" check is `git diff --cached` against the local copy, not the remote. After a failed rebase the local branch still holds the un-pushed commit (which contains the desired files), so the next reconcile sees "no changes", returns success, and marks the WorkPlacement `Ready`. Status ends up flipping between `Failing` (on a real change) and `Ready` (on the next no-op reconcile), even though the repo never received the files.

Suggested fix: before writing, refresh the local copy against the remote instead of only clearing uncommitted changes — abort any in-progress rebase, `git fetch` the branch (shallow, to keep PR #726's performance win), then `git reset --hard origin/<branch>` before `git clean -ffdx`. That removes the divergent local commit (so the push is a clean fast-forward) and makes the change-detection compare against the real remote (so a deleted file is re-written and status reflects reality). `Reset()` already runs at the start of every write in both the WorkPlacement and Destination controllers, so one change covers both paths.

Reproduce: write a resource to a Git state store, delete its file directly on the remote branch, then change the resource so it gets written again.

### Acceptance criteria

```gherkin
Scenario: Recover when the repo is changed outside Kratix
  Given Kratix has written a resource to a Git state store
  And that resource's file is deleted directly on the remote branch
  When the resource is written again
  Then the write succeeds and the file is back on the remote
  And other resources using the same state store keep writing normally

Scenario: Status reflects what is really in the repo
  Given a resource's file is missing from the remote branch
  When its WorkPlacement is reconciled
  Then it is only Ready once the remote actually contains the file
```

### Tasks

- [ ] Refresh the local clone against the remote before each write (clear any in-progress rebase, fetch, hard reset)
- [ ] Make write success and `Ready` reflect the remote, not the local copy
- [ ] Add tests: recovers after the repo is changed externally; not `Ready` while the remote is missing files

aclevername:
closes #850

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -64,6 +64,7 @@ import (
 	"github.com/syntasso/kratix/internal/logging"
 	"github.com/syntasso/kratix/internal/telemetry"
 	"github.com/syntasso/kratix/lib/fetchers"
+	gitutil "github.com/syntasso/kratix/util/git"
 	//+kubebuilder:scaffold:imports
 )
 
@@ -83,6 +84,7 @@ type KratixConfig struct {
 	Telemetry                      *telemetry.Config             `json:"telemetry,omitempty"`
 	Logging                        *LoggingConfig                `json:"logging,omitempty"`
 	ResourceBindingVersionStrategy ResourceBindingDefaultVersion `json:"resourceBindingVersionStrategy,omitempty"`
+	Git                            *GitConfig                    `json:"git,omitempty"`
 }
 
 // ResourceBindingDefaultVersion controls the version strategy for ResourceBindings.
@@ -98,6 +100,10 @@ const (
 type LoggingConfig struct {
 	Structured *bool   `json:"structured,omitempty"`
 	Level      *string `json:"level,omitempty"`
+}
+
+type GitConfig struct {
+	MinimumFetchInterval *metav1.Duration `json:"minimumFetchInterval,omitempty"`
 }
 
 type Workflows struct {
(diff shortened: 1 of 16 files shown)


aclevername:
@codex review

PR #856, opened 2026-07-27
feat: use PromiseRevision for requiredPromises

### LINKED ISSUE #832: feat: requiredPromises use PromiseRevisions
**As a** Promise writer
**I want** compound promises to respect promise revisions
**So that** I don't have to upgrade my compound promise every time a new version of the sub-promise changes

This came as a surprise to customers that it doesn't happen already. 
When I reference a requiredPromise in a compound, it should use the PromiseRevision rather than reference the promise directly (like how resources can do via the resourcebindings).

This means that:
1. Compounds don't become unavailable when one of their requiredPromises is updated
2. Compound promise authors aren't tied to the upgrade flow of the promises they depend on, meaning they have greater control over their user experience

## Tech notes
This is complicated due to new requests being served at latest version (even if using the string, not latest tag). So if a user has requiredPromise at v2, and the latest is v3, when you create a new request to the Promise you would send a request to v3 which is potentially confusing (as you would expect v2).

Option: Raise the resource binding before the request, then the request would be to v2
**Decision:** Do this for the promise and document that users should apply the resource binding before sending the request

## Acceptance

```gherkin
Given two PromiseRevisions - redis-v1 and redis-v2
When I install an app Promise that defines redis at v1 as a required Promise
Then the Promise in installed and advertised as 'Available'
And the RequirementsFulfilled condition is 'true;
```

ChunyiLyu:
## Context

closes #832

This PR includes 2 feature changes:
1. use promise revision to resolve required promises for compounds
2. support initial version pin of requests by using label to find resource bindings first before trying to create it

* When the required promise version is not the latest promise version, compound promise stays available as long as the revision is present:
```
❯ k get promiserevisions
NAME                PROMISE       VERSION   LATEST
widget-081a7        widget        v2.0.0    true
widget-2888c        widget        v1.0.0
```

Compound Promise
```
...
  requiredPromises:
  - name: widget
    version: v1.0.0
  kind: App
...
  lastAvailableTime: "2026-07-27T13:59:38Z"
  observedGeneration: 2
  requiredPromises:
  - name: widget
    state: Requirement installed
    version: v1.0.0
```

* When required promise installed, but the required version cannot be found:
```
❯ kubectl apply -f compound-promise.yaml
Warning: Required Promise "widget" installed, but no revision found at version "v1.0.0"
Warning: Promise will not be available until the above issue(s) is resolved
promise.platform.kratix.io/app created
...
  kind: App
  kratix:
    apiVersion: example.promise.syntasso.io/v1
    kind: App
    status: Unavailable
    version: v1.0.0
    workflows: {}
  requiredPromises:
  - name: widget
    state: Requirement not installed at the specified version
    version: v1.0.0
  status: Unavailable
```

### DIFF
--- a/internal/webhook/v1alpha1/promise_webhook.go
+++ b/internal/webhook/v1alpha1/promise_webhook.go
@@ -237,8 +237,8 @@ func validateRequiredPromisesAreAvailable(p *v1alpha1.Promise) admission.Warning
 		}
 		if !satisfied {
 			warnings = append(warnings, fmt.Sprintf(
-				"Required Promise %q installed, but no revision found at version %q; it is currently at %q",
-				requirement.Name, requirement.Version, promise.Status.Version))
+				"Required Promise %q installed, but no revision found at version %q",
+				requirement.Name, requirement.Version))
 		}
 	}
 
--- a/internal/webhook/v1alpha1/promise_webhook.go
+++ b/internal/webhook/v1alpha1/promise_webhook.go
@@ -226,8 +226,19 @@ func validateRequiredPromisesAreAvailable(p *v1alpha1.Promise) admission.Warning
 			promiselog.Error(err, "failed to get requirement", "requirement", requirement.Name, "version", requirement.Version)
 			continue
 		}
-		if promise.Status.Version != requirement.Version {
-			warnings = append(warnings, fmt.Sprintf("Required Promise %q installed but not at a compatible version, want: %q have: %q", requirement.Name, requirement.Version, promise.Status.Version))
+		// Requirements are satisfied by the PromiseRevisions a Promise has left behind, not
+		// by the version it happens to be running, so a Promise upgraded past the required
+		// version still satisfies it.
+		satisfied, err := requiredRevisionExists(requirement, promise)
+		if err != nil {
+			promiselog.Error(err, "failed to look up revisions for requirement",
+				"requirement", requirement.Name, "version", requirement.Version)
+			continue
+		}
+		if !satisfied {
+			warnings = append(warnings, fmt.Sprintf(
+				"Required Promise %q installed, but no revision found at version %q; it is currently at %q",
+				requirement.Name, requirement.Version, promise.Status.Version))
 		}
 	}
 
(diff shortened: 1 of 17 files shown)


internal/webhook/v1alpha1/promise_webhook.go
```diff
@@ -237,6 +248,30 @@ func validateRequiredPromisesAreAvailable(p *v1alpha1.Promise) admission.Warning
 	return warnings
 }
 
+// requiredRevisionExists reports whether the required Promise has a PromiseRevision
+// satisfying the requirement. Mirrors the controller's rule so an admission warning cannot
+// contradict the status the controller will write moments later.
+func requiredRevisionExists(requirement v1alpha1.RequiredPromise, required *v1alpha1.Promise) (bool, error) {
```
shano, line 254: This looks like a bit of a divergent but duplicated implementation of "does a revision exist" compared to the one in the promise_controller, it differs in that in the controller it'll treat two PromiseRevisions matching as an error, where the webhook here has no duplicate checks at all. Not blocking but might be worth a shared exported helper, so both usages are consistent. I am guessing the webhook will silently allow this but the controller will flag as unknown(I guess)?

shano (approved):
Tested locally:

- Reproduced PR scenario: created sub-promise@v1.0.0, a compound requiring it → Available. Upgraded sub-promise to v2.0.0 → compound stayed Available for 40s straight, still pinned to v1.0.0
- Triggered the missing-version case (requiredPromise@v9.9.9, no matching revision) → got the exact warning text from the PR description, promise went Unavailable.

internal/webhook/v1alpha1/promise_webhook.go
ChunyiLyu: So far all revisions are created by kratix and there won't be any duplicates. It could happen when users hand create promise revisions but I think more logic breaks if there are repeated revisions (like resource request reconciliation would fail). 

You're right that they diverge, however to me it would be weird to surface this error to our users at webhook level because this is not a result of them installing this compound promise. It's an operational error from something else.

PR #860, opened 2026-07-30
feat: expose configuring workplacement drift detection minimum resync…

aclevername:
… time

### DIFF
--- a/internal/controller/workplacement_controller.go
+++ b/internal/controller/workplacement_controller.go
@@ -68,10 +68,10 @@ type WorkPlacementReconciler struct {
 
 	RepositoryCache RepositoryCache
 
-	// RewriteInterval is how often a WorkPlacement re-writes its files to the
-	// state store even when nothing has changed. When zero, no periodic re-write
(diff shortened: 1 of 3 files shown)


internal/controller/workplacement_controller.go
```diff
@@ -82,6 +88,8 @@ type workPlacementReconcileContext struct {
 	repositoryCache RepositoryCache
 
 	versionCache map[string]string
+
+	reconciliationInterval time.Duration
```
ChunyiLyu: nitpick: as a variable it's the same as the general `reconciliationinterval` which was a bit confusing to look at. Maybe can rename to `rewriteInterval` or something? WDT?

aclevername: agreed, pushed a fix

ChunyiLyu (approved):
Nice! `StateStoreReconciliationInterval` is a good name 👍

aclevername:
docs https://github.com/syntasso/kratix-docs/pull/597

PR #866, opened 2026-08-05
feat: reconcile promise workflow after failure

### LINKED ISSUE #862: feat: Reconcile promises on the schedule even when the previous workflow run failed
## Context
Kratix is intended to continuously reconcile: workflows should re-run on the Default Reconciliation Interval the same way a Kubernetes controller keeps driving towards desired state. Today the periodic reconciliation only re-runs a Promise's workflows when they all last completed successfully. A Promise whose workflow last failed is skipped, so it sits in a failed state and Kratix never retries it on its own.

The only ways to get it going again today are:
- change the Promise spec
- add the `kratix.io/manual-reconciliation: "true"` label

This story makes the periodic reconciliation re-run Promise workflows regardless of the previous run's outcome, and adds a Kratix config field so operators can opt into the alternative (do not re-run after a failure). The default is continuous reconciliation, including retry after a failed run.

This mirrors #858, which covers the same behaviour for Resource workflows.

## What to build
By default, the periodic reconciliation re-runs a Promise's workflows on the schedule whether the previous run succeeded or failed. A failed run no longer stops future scheduled runs — Kratix keeps reconciling towards desired state.

A new field in the Kratix config lets an operator disable this, so a failed run is skipped by the periodic reconcile until it is manually addressed. The default value keeps continuous reconciliation on, including retry after failure. The field is documented so operators understand the trade-off.

## Engineering notes
- This should share the same Kratix Config field introduced in #858 (`workflows.reconcileAfterFailure`, default `true`) rather than introducing a separate Promise-specific field, so behaviour is consistent across Promises and Resources. Wire it through to the Promise periodic reconcile path.
- The relevant guard today skips Promises whose workflows did not all last complete successfully. That is the behaviour this field gates. Suspended/paused handling is separate and out of scope here.
- Default must preserve continuous reconciliation, including re-running after a failed run. Only when an operator explicitly disables it should a failed run be skipped by the periodic reconcile.

## Acceptance criteria

```gherkin
Given a Promise whose previous workflow run failed
And the Kratix config uses the default settings
When the Default Reconciliation Interval elapses
Then the Promise's workflows are re-run

Given a Promise whose previous workflow run succeeded
And the Kratix config uses the default settings
When the Default Reconciliation Interval elapses
Then the Promise's workflows are re-run

Given a Promise whose previous workflow run failed
And the operator has disabled retry-after-failure in the Kratix config
When the Default Reconciliation Interval elapses
Then the Promise's workflows are not re-run
And the Promise can still be reconciled manually via a spec change or the kratix.io/manual-reconciliation label
```

jeduardo:
Re-runs failed Promise configure workflows on the normal reconciliation schedule unless `workflows.reconcileAfterFailure` is disabled. Adds coverage for retries, job cleanup, recovery, and manual reconciliation.

Closes #862


config/samples/kratix-config.yaml
```diff
@@ -11,6 +11,9 @@ data:
       defaultContainerSecurityContext:
         windowsOptions:
           runAsUserName: "setInKratixConfig"
+      # reconcileAfterFailure controls whether the periodic reconciliation (on the
+      # Default Reconciliation Interval) re-runs a Promise's or resource request's
+      # configure workflows after the previous run failed. Defaults to true.
```
jeduardo: These comments can be dropped.

cmd/main.go
```diff
@@ -121,7 +121,8 @@ type Workflows struct {
 	DefaultImagePullPolicy          corev1.PullPolicy            `json:"defaultImagePullPolicy,omitempty"`
 	DefaultContainerResources       *corev1.ResourceRequirements `json:"defaultContainerResources,omitempty"`
 	JobOptions                      JobOptions                   `json:"jobOptions,omitempty"`
-	// Controls whether the periodic reconciliation re-runs resource workflows after they failed. Defaults to true
+	// ReconcileAfterFailure controls whether the periodic reconciliation re-runs a Promise's
+	// or resource request's configure workflows after the previous run failed. Defaults to true
```
jeduardo: These comments can be dropped

docs/superpowers/specs/2026-08-05-reconcile-promise-workflow-after-failure-design.md
```diff
@@ -0,0 +1,225 @@
+# Reconcile Promise workflows after failure (#862)
```
jeduardo: Superpowers plans can be dropped

PR #871, opened 2026-08-12
chore: migrate quickstart to use seaweedfs instead of minio

senglezou:
Replaces MinIO with SeaweedFS as the local S3 statestore used by the quickstart, the scripts and the system tests. MinIO Community Edition is unmaintained and source-only, and the `mc` client the Makefile fetched comes from the commercial AIStor distribution.

## What changed

**Statestore.** `config/samples/minio-install.yaml` becomes `config/samples/seaweedfs-install.yaml`, pinned to `chrislusf/seaweedfs:4.41` (`hack/platform/` keeps its symlink). The S3 Service keeps nodePort `31337` but is renamed `bucket`: the DNS name is a contract other repos hardcode, so it stays vendor-neutral. Credentials live in a `bucket-credentials` secret and are read by the server from the environment.

**Bucket creation** is an ordinary second container in the SeaweedFS pod rather than a one-shot Job. The data volume is an `emptyDir`, so a pod replacement loses the bucket and something has to put it back; the Job could not, and was deleted at the end of install. A plain container rather than a native sidecar keeps the manifest free of a Kubernetes 1.29 floor. Readiness checks that the bucket exists, not just that the port is open, because `quick-start.sh` and `helm-e2e-test.sh` both gate on `condition=ready` and mean "the state store is usable".

**Bucket CLI.** `make s5cmd-cli` installs `bin/s5cmd` in place of the `mc` binary. It is a wrapper that reads the endpoint and credentials from the cluster, so `bin/s5cmd ls` works with nothing exported, and `make bucket-env` prints the same settings for other S3 tools. With no local bucket it refuses rather than falling through to the default AWS endpoint, where ambient credentials could point an `rm` at a real account.

**Tests and waits.** The system tests read the state store over S3 instead of shelling out to a CLI. The reconcile wait is bounded on wall-clock time and dumps diagnostics on failure, so a wedged install fails with output instead of hanging until the job is cancelled. `run` streams output in CI rather than buffering it.

**Docs.** The `docs/deprecated` tree is deleted rather than migrated: it described a setup that no longer matches the scripts.

No change to `lib/writers/s3.go` beyond comments: the `minio-go` client stays, and SeaweedFS needs no addressing or S3-semantics changes.

## Verification

- `make quick-start` end to end, then `bin/s5cmd ls` / `ls s3://kratix/*` / `cat` against the running bucket
- the CLI wrapper: fresh install, no-cluster refusal, `version`/`--help` without a cluster, explicit `S3_ENDPOINT_URL` override, missing-binary path


config/samples/platform_v1alpha1_bucketstatestore.yaml
```diff
@@ -3,10 +3,10 @@ kind: BucketStateStore
 metadata:
   name: default
 spec:
-  endpoint: minio.kratix-platform-system.svc.cluster.local
+  endpoint: seaweedfs.kratix-platform-system.svc.cluster.local
```
senglezou: We should have the bucket endpoint with a more vendor-agnostic name such as `bucket` instead of seaweedfs

jeduardo:
@codex review

hack/destination/gitops-tk-resources-single-cluster.yaml
```diff
@@ -8,20 +8,20 @@ spec:
   interval: 10s
   provider: generic
   bucketName: kratix
-  endpoint: minio.kratix-platform-system.svc.cluster.local
+  endpoint: seaweedfs.kratix-platform-system.svc.cluster.local
```
senglezou:
```suggestion
  endpoint: bucket.kratix-platform-system.svc.cluster.local
```

hack/platform/helm-values-bucket.yaml
```diff
@@ -0,0 +1,12 @@
...
+- kind: BucketStateStore
+  name: default
+  namespace: default
+  secretRef:
+    name: bucket-credentials
+    values:
+      accesskey: a3JhdGl4YWRtaW4=
+      secretkey: a3JhdGl4YWRtaW4=
+  insecure: true
+  endpoint: seaweedfs.kratix-platform-system.svc.cluster.local
```
senglezou:
```suggestion
  endpoint: bucket.kratix-platform-system.svc.cluster.local
```

scripts/quick-start.sh
```diff
@@ -238,9 +238,8 @@ setup_platform_destination() {
         kubectl --context kind-${PLATFORM_CLUSTER_NAME} apply --filename "${ROOT}/hack/platform/gitea-install.yaml"
     fi
 
-    if ${INSTALL_AND_CREATE_MINIO_BUCKET}; then
-        make minio-cli
```
senglezou, line 247: double check where the binary is coming from now

scripts/quick-start.sh
```diff
@@ -293,21 +292,14 @@ wait_for_gitea() {
     kubectl wait job --context kind-${PLATFORM_CLUSTER_NAME} -n gitea gitea-create-repository --for condition=Complete ${wait_opts}
 }
 
-wait_for_minio() {
+wait_for_seaweedfs() {
     wait_opts=$1
-    while ! kubectl get pods --context kind-${PLATFORM_CLUSTER_NAME} -n kratix-platform-system | grep minio; do
+    while ! kubectl get pods --context kind-${PLATFORM_CLUSTER_NAME} -n kratix-platform-system | grep seaweedfs; do
```
senglezou: we could do a label selector.

scripts/quick-start.sh
```diff
@@ -321,20 +313,46 @@ wait_for_local_repository() {
...
+
+# The first call is given a short budget and its failure is recoverable: the
+# caller re-runs the worker setup and calls again with --no-timeout. That second
+# call is still bounded, so a wedged install fails with diagnostics rather than
+# hanging until the CI job is cancelled hours later.
 wait_for_namespace() {
     local timeout_flag="${1:-""}"
+    local max_loops=60
+    if [ -n "${timeout_flag}" ]; then
+        max_loops=300
```
senglezou: 5 hours timeout in total is a lot, we should reduce it to an hour

test/system/assets/destination/destination-test-store.yaml
```diff
@@ -5,8 +5,8 @@ metadata:
 spec:
   authMethod: accessKey
   bucketName: kratix
-  endpoint: minio.kratix-platform-system.svc.cluster.local
+  endpoint: seaweedfs.kratix-platform-system.svc.cluster.local
```
senglezou:
```suggestion
  endpoint: bucket.kratix-platform-system.svc.cluster.local
```

Issue #870, opened 2026-08-12
feat: Ability to set reconciliation on a per-Promise basis


catmo-syntasso:
**As a platform engineer**
**I want** to configure reconciliation intervals on a per-promise basis
**So that** I can tune reconciliation to API rate limits

## Context
Customers have requested the ability to configure how often a promise reconciles. The example provided was in response to API rate limits. This is also valuable more generically for if you have any promise that should reconcile at a different interval than the default. This also allows Promise authors some control to the drift detection intervals in their promise.

As the variable should be set on a per-promise basis, it needs to form part of the Promise definition, but this story does not determine what this should be. A label or annotation seem most obvious but other options should be explored in this story.

Consideration needs to be given to resource spikes if the interval is set too often.

## Engineering notes

* This value will takes precedence over the `defaultReconciliationInterval`

## Acceptance
```gherkin
Scenario: Promise has reconciliation interval set
Given a Promise has the reconciliation interval set
And it is different to the default interval
When that duration has elapsed
Then the promise and corresponding resource pipelines reconcile
And none of the other Promises reconcile any pipelines

Scenario: No reconciliation interval set
Given a Promise does *not* have a reconciliation interval set
Then it reconciles at the default platform-configured interval
```

kirederik:
## Design: where the reconciliation interval lives

The story asked whether this should be a label or part of the Promise spec. The answer is **both, at different layers** — because there are two owners with different needs, and the codebase treats them differently.

**Vanessa (Promise author)** knows her pipeline hits a rate-limited vendor API. That's domain knowledge that should travel with the artifact and be versioned → **spec**.

**Penny (platform engineer)** tunes an installed Promise on her own cluster. A spec edit is reverted on the next `PromiseRelease` resync (`existingPromise.Spec = promise.Spec`, promiserelease_controller.go:180), so she needs a path that survives → **annotation**.

### Who writes where

```mermaid
flowchart LR
    V["Vanessa<br/>Promise author"] -->|"spec.workflows.config"| P[Promise]
    Pe["Penny<br/>Platform engineer"] -->|"annotation"| P
    Pe -->|"annotation, per version"| R2["PromiseRevision v1.0.0<br/>older, pinned"]
    P -->|"spec snapshot<br/>+ annotation sync"| R1["PromiseRevision v2.0.0<br/>latest"]
    R1 --> W1["Promise workflows<br/>+ resources on v2.0.0"]
    R2 --> W2["Resources pinned to v1.0.0"]
[...]

PR #877, opened 2026-08-17
feat: per-Promise reconciliation interval

kirederik:
Closes part of #870. **First of four stacked PRs — review in order:**

| PR | |
|---|---|
| **#877** (this) | A Promise author declares an interval in `spec` |
| #878 | A platform engineer overrides it per revision |
| #880 | …or once, on the Promise |
| #881 | Removes the delivery plans |

---

A Promise author can say how often their Promise's workflows re-run:

```yaml
spec:
  workflows:
    config:
      reconciliationInterval: 30m
```

Unset, it falls back to the platform-wide `reconciliationInterval`, then to the 10h default. It applies to the Promise's own workflows and to every resource request, and it governs failure retries as well as successful re-runs.

### It resolves from the PromiseRevision, not the live Promise

Each object reads the interval from the revision it's bound to, so a resource pinned to v1.0.0 keeps that version's cadence after v2.0.0 ships. This also means changing the interval means cutting a version — which is what you want, because the promise hash feeds the resource pipeline hash, so a spec edit re-runs every workflow anyway.

### Validation

Zero, negative and sub-minute values are rejected on create and update. Zero is rejected rather than meaning "never", because the force-run gate compares `time.Since(...) <= interval` — at zero that's always false, so it would fire on *every* reconcile. Use `kratix.io/paused` to stop reconciliation.

The read path clamps too. `PromiseRevision` has no validating webhook and a bare string schema, so the floor was otherwise a property of one write path rather than a guarantee at the point of use.

`kratix.io/paused` still wins outright, `kratix.io/manual-reconciliation` still bypasses the interval, and delete workflows are unaffected.

### Known limitations

- **An operator can't cap what an author chooses.** The one-minute floor is a compile-time constant while the platform default is configurable, and Promise spec arrives verbatim from an OCI or git artefact.
- **Failure retries are flat at the interval, no backoff.** At the floor that's 60 retries an hour against an API that may already be throttling.
- **No ceiling.** `8760h` is accepted and effectively stops periodic re-runs.
- **The effective interval appears only in logs** — no status field or printer column.

### Verification

`make lint` 0 issues, `make test` 15 suites.


api/v1alpha1/promiserevision_types.go
```diff
@@ -145,6 +147,28 @@ func (pr *PromiseRevision) ClearLatestRevisionLabel() {
...
 }
 
+// MinReconciliationInterval is the floor the Promise validating webhook enforces for
+// spec.workflows.config.reconciliationInterval. ReconciliationInterval enforces the same floor
+// on a revision's snapshot.
+const MinReconciliationInterval = time.Minute
+
+// ReconciliationInterval returns this revision's snapshot reconciliation interval, or fallback
+// if the snapshot does not set one or sets one below MinReconciliationInterval.
+func (pr *PromiseRevision) ReconciliationInterval(fallback time.Duration) time.Duration {
```
aclevername: wonder if we can just hardcode the fallback, simplify this a bit

api/v1alpha1/promiserevision_types_test.go
```diff
@@ -73,4 +75,82 @@ var _ = Describe("PromiseRevision", func() {
...
+						},
+					},
+				},
+			}
+			revision := platformv1alpha1.NewPromiseRevision(promise, "v1.0.0")
+
+			promise.Spec.Workflows.Config.ReconciliationInterval = &metav1.Duration{Duration: 99 * time.Minute}
+
+			Expect(revision.ReconciliationInterval(fallback)).To(Equal(3 * time.Minute))
+		})
```
aclevername: this just seems silly, delete

api/v1alpha1/promiserevision_types_test.go
```diff
@@ -73,4 +75,82 @@ var _ = Describe("PromiseRevision", func() {
...
+			revision := revisionWithInterval(platformv1alpha1.MinReconciliationInterval)
+			Expect(revision.ReconciliationInterval(fallback)).To(Equal(platformv1alpha1.MinReconciliationInterval))
+			Expect(revision.ReconciliationIntervalBelowMinimum()).To(BeFalse())
+		})
+
+		It("does not report unset as below minimum", func() {
+			revision := &platformv1alpha1.PromiseRevision{}
+			Expect(revision.ReconciliationIntervalBelowMinimum()).To(BeFalse())
+		})
+	})
```
aclevername: nit: having a describe table when its just 2 tests and rest of suite is 3 Its seems odd. rather just have a bit of repetition

also what does "clamp" mean

config/crd/bases/platform.kratix.io_promises.yaml
```diff
@@ -117,6 +117,11 @@ spec:
                           resource) to run in; needs to be an existing namespace or
                           else workflow won't start
                         type: string
+                      reconciliationInterval:
+                        description: Interval at which all workflows (promise and
+                          resource) defined in this Promise are reconciled; if not
+                          set, defaults to the platform-wide reconciliation interval
+                        type: string
```
aclevername, line 124: technically its only for Promise workflow if the reivsion is latest. minor nit and not worth fixing since its just mirroring from the promise CRD

internal/controller/dynamic_resource_request_controller.go
```diff
@@ -1462,9 +1463,18 @@ func workflowInProgress(workflowCompletedCondition *clusterv1.Condition) bool {
...
+	interval := promiseRevisionUsed.ReconciliationInterval(r.ReconciliationInterval)
+	source := "revision"
+	if promiseRevisionUsed.Spec.PromiseSpec.Workflows.Config.ReconciliationInterval == nil {
+		source = "globalDefault"
+	} else if promiseRevisionUsed.ReconciliationIntervalBelowMinimum() {
+		source = "globalDefault"
+		logging.Warn(logger, "PromiseRevision reconciliation interval below minimum; using fallback",
+			"promiseRevision", promiseRevisionUsed.GetName(), "minimum", v1alpha1.MinReconciliationInterval)
+	}
+	logging.Info(logger, "scheduling next reconciliation", "reconciliationInterval", interval, "source", source)
```
aclevername: having the helper method but then still reading the value to set the `source` feels weird. helper should just do that too, or remove the helper

internal/controller/dynamic_resource_request_controller_test.go
```diff
@@ -599,6 +600,228 @@ var _ = Describe("DynamicResourceRequestController", func() {
...
+					Expect(err).NotTo(HaveOccurred())
+					Expect(result).To(Equal(ctrl.Result{}))
+
+					Expect(fakeK8sClient.Get(ctx, nn, resReq)).To(Succeed())
+					Expect(resReq.GetLabels()[resourceutil.ManualReconciliationLabel]).NotTo(Equal("true"))
+				})
+			})
+		})
+	})
+
```
aclevername, line 692: 220 lines of tests for this feature change seems excessive. also hard to read. i think we can prompt claude to refactor this

internal/controller/promise_controller.go
```diff
@@ -676,9 +676,36 @@ func (r *PromiseReconciler) reconcileResources(ctx context.Context, logger logr.
...
+
+// resolveReconciliationInterval resolves the reconciliation interval for promise and reports
+// which tier supplied it: the interval declared on promise's latest PromiseRevision ("revision");
+// falling back to the interval declared on the live Promise spec when no revision is marked
+// latest yet ("promiseSpec"); falling back to r.ReconciliationInterval when neither declares one,
+// or when the resolved revision's interval is below the minimum ("globalDefault").
+func (r *PromiseReconciler) resolveReconciliationInterval(ctx context.Context, promise *v1alpha1.Promise, logger logr.Logger) (time.Duration, string) {
+	revision, err := latestRevision(ctx, r.Client, promise)
+	if err == nil {
+		if revision.ReconciliationIntervalBelowMinimum() {
```
aclevername: should we juat validating webhook this, then we can remove all the worry about this in the code?

internal/controller/promise_controller_test.go
```diff
@@ -2157,6 +2157,133 @@ var _ = Describe("PromiseController", func() {
...
+				Expect(fakeK8sClient.Create(ctx, otherPromise)).To(Succeed())
+				otherPromiseName := client.ObjectKeyFromObject(otherPromise)
+				Expect(fakeK8sClient.Get(ctx, otherPromiseName, otherPromise)).To(Succeed())
+				otherPromise.UID = "5678efgh"
+				Expect(fakeK8sClient.Update(ctx, otherPromise)).To(Succeed())
+
+				result, err := t.reconcileUntilCompletion(reconciler, otherPromise)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(result).To(Equal(ctrl.Result{RequeueAfter: reconciler.ReconciliationInterval}))
+			})
```
aclevername: overly defensive

internal/webhook/v1alpha1/promise_webhook.go
```diff
@@ -104,9 +104,26 @@ func validatePromise(p *v1alpha1.Promise) ([]string, error) {
...
+	interval := p.Spec.Workflows.Config.ReconciliationInterval
+	if interval == nil {
+		return nil
+	}
+	if interval.Duration < v1alpha1.MinReconciliationInterval {
+		return fmt.Errorf("spec.workflows.config.reconciliationInterval: Invalid value: %q: must be at least %s",
+			interval.Duration, v1alpha1.MinReconciliationInterval)
+	}
+	return nil
+}
```
aclevername, line 125: oh we are validating it ffs.

internal/controller/promise_controller.go
```diff
@@ -676,9 +676,36 @@ func (r *PromiseReconciler) reconcileResources(ctx context.Context, logger logr.
...
+	logging.Info(logger, "scheduling next reconciliation", "reconciliationInterval", interval, "source", source)
+	return ctrl.Result{RequeueAfter: interval}
+}
+
+// resolveReconciliationInterval resolves the reconciliation interval for promise and reports
+// which tier supplied it: the interval declared on promise's latest PromiseRevision ("revision");
+// falling back to the interval declared on the live Promise spec when no revision is marked
+// latest yet ("promiseSpec"); falling back to r.ReconciliationInterval when neither declares one,
+// or when the resolved revision's interval is below the minimum ("globalDefault").
+func (r *PromiseReconciler) resolveReconciliationInterval(ctx context.Context, promise *v1alpha1.Promise, logger logr.Logger) (time.Duration, string) {
```
aclevername, line 690: this can probs be condensed

internal/controller/promise_controller.go
aclevername: turns out we already are, i'd be in favour of removing all these redundant checks

internal/controller/promise_controller_test.go
```diff
@@ -2157,6 +2157,133 @@ var _ = Describe("PromiseController", func() {
...
+
+			It("forces a re-run", func() {
+				result, err := reconciler.Reconcile(ctx, ctrl.Request{NamespacedName: promiseName})
+				Expect(err).NotTo(HaveOccurred())
+				Expect(result).To(Equal(ctrl.Result{}))
+
+				Expect(fakeK8sClient.Get(ctx, promiseName, promise)).To(Succeed())
+				Expect(promise.Labels[resourceutil.ManualReconciliationLabel]).To(Equal("true"))
+			})
+		})
```
aclevername, line 2255: i don't quite get what logic its actually testing, i can't see an impl change that affects this? might be redundant.

internal/controller/promise_controller.go
kirederik: webhooks can be disabled though

api/v1alpha1/promiserevision_types_test.go
aclevername: Deleted.

api/v1alpha1/promiserevision_types_test.go
aclevername: Dropped the table, back to two plain Its like the rest of the suite. On "clamp": it meant flooring a sub-minimum interval up to the fallback. That behaviour is gone now (the webhook enforces the floor), so the wording went with it.

internal/controller/dynamic_resource_request_controller.go
aclevername: The helper now returns `(interval, bool)` where the bool is just "did the author declare one". Each controller names its own source ("revision"/"globalDefault"), so we no longer re-read the field to work it out, and the log vocabulary stays out of the api type.

internal/controller/dynamic_resource_request_controller_test.go
aclevername: Trimmed. Removed the cross-promise isolation test (it spun up a whole second reconciler, which was the worst of the noise) and the now-dead below-minimum machinery. Block is roughly halved.

internal/controller/promise_controller.go
aclevername: Agreed. The validating webhook is the single enforcement point now, so I removed the below-minimum checks here and the read-path clamp entirely.

internal/controller/promise_controller_test.go
aclevername: Deleted.

internal/webhook/v1alpha1/promise_webhook.go
aclevername: Yep. Leaned on it as the single source of truth and removed the redundant downstream checks plus the read-path clamp.

internal/controller/promise_controller.go
aclevername: Condensed. The below-minimum branch is gone, so it is just the three tiers now.

aclevername (approved):
lgtm

PR #878, opened 2026-08-17
feat: PromiseRevision annotation overrides the reconciliation interval

kirederik:
**Stacked on #877.** Second of four — see #877 for the stack map.

Slice 1 let a Promise *author* declare an interval. This lets a platform engineer override it on a Promise she doesn't own, without forking it:

```bash
kubectl annotate promiserevision postgres-1-0-0 kratix.io/reconciliation-interval=30m
```

An annotation on the revision beats the spec value inside it. It works where a spec edit doesn't, because `PromiseRelease` overwrites `spec` on every resync but merges annotations.

### Validation

Unparseable and sub-minute values are rejected on create and update, with distinct messages. The read path declines rather than rejects, falling through to the spec snapshot. Admission only validates when the value *changes*, so raising the floor later can't lock an existing revision out of every future update.

### Propagation

Changing the annotation enqueues the owning Promise, so its own workflows pick up the new cadence promptly. **Resource requests are not fanned out** — they pick it up at their next reconcile, or immediately if you set `kratix.io/reconcile-resources` on the Promise.

The watch predicate compares resolved durations, so reformatting `60s` to `1m0s` isn't a change.

### Security fix included

The interval-driven force-run ran *before* the suspension check and wrote `kratix.io/manual-reconciliation`, which the suspension handler read as consent — **deleting** `kratix.io/workflow-suspended` rather than bypassing it. Combined with this slice, patching a PromiseRevision could permanently un-suspend a Promise and every resource request bound to it, in namespaces you otherwise can't touch.

Fixed by guarding the gate rather than the handler, so a human setting `manual-reconciliation` still resumes a suspended workflow as [ADR0008](https://github.com/syntasso/adrs/blob/main/0008_workflow_controls.md) documents. Both behaviours are asserted.

### Also closes slice 1's outstanding test gap

Slice 1 couldn't prove its controllers read from the revision rather than the Promise, because `promise.Spec` is assigned from the revision snapshot before every call site. The annotation lives in metadata, which that assignment never copies — so the discriminating test is now real.

### Known limitations

Unchanged from #877: no ceiling, no status surface, floor not operator-configurable, retries flat at the interval.

### Verification

`make lint` 0 issues, `make test` 15 suites, 389 controller specs.

### DIFF
--- a/api/v1alpha1/promiserevision_types.go
+++ b/api/v1alpha1/promiserevision_types.go
@@ -147,18 +147,42 @@ func (pr *PromiseRevision) ClearLatestRevisionLabel() {
 	delete(l, LatestRevisionLabel)
 }
 
-// MinReconciliationInterval is the floor the Promise validating webhook enforces for
-// spec.workflows.config.reconciliationInterval.
+// MinReconciliationInterval is the floor enforced by the Promise validating webhook for
+// spec.workflows.config.reconciliationInterval, by the PromiseRevision validating webhook for
+// ReconciliationIntervalAnnotation, and by ReconciliationInterval when it reads that annotation.
 const MinReconciliationInterval = time.Minute
 
-// ReconciliationInterval returns this revision's snapshot reconciliation interval and true, or
-// fallback and false when the snapshot does not declare one.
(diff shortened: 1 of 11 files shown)


aclevername (approved):
lgtm

PR #880, opened 2026-08-17
feat: set the reconciliation interval once on the Promise

kirederik:
**Stacked on #878.** Third of four — see #877 for the stack map. This completes the feature.

Set the interval once on the Promise instead of on every revision:

```bash
kubectl annotate promise postgres kratix.io/reconciliation-interval=30m
```

The revision for the Promise's current version mirrors it, and loses it when you remove it from the Promise. Revisions for older versions keep whatever they carry, so resources pinned to an old version stay individually tunable.

### How the interval resolves, with all three slices in

Every object reads from the PromiseRevision it is bound to:

```mermaid
flowchart TD
    R["The PromiseRevision this object is bound to"] --> A
    A{"annotation set,<br/>parses, ≥ 1m?"} -->|yes| U1["that value"]
    A -->|no| S{"spec snapshot<br/>set, ≥ 1m?"}
    S -->|yes| U2["that value"]
    S -->|no| U3["platform default"]
```

Two things can put that annotation on a revision: an operator, by hand on any revision; or this PR's mirror, copying it from the Promise onto the current version's revision.

### PromiseRelease-managed Promises are not mirrored

For those, `installPromise` merges the artefact's annotations over the live Promise on every resync — so the value would be the artefact author's, not yours, and your edit would be reverted by the reconcile it triggers. Annotate the revision directly instead; that path is deliberately left working for them.

### Validation

Admission rejects unparseable and sub-minute values on the Promise, sharing one validator with the PromiseRevision webhook. An existing out-of-policy value is grandfathered on update, so raising the floor later can't lock a Promise out of every edit. The mirror declines a bad value rather than propagating it into a write admission would reject.

### Known limitations

- **An override can be stranded on an archived revision.** Set the annotation, ship a new Promise version, then remove it: resources pinned to the old version keep the old interval, and nothing surfaces that. Doesn't affect PromiseRelease-managed Promises, which aren't mirrored.
- **No ceiling.** `8760h` is accepted and effectively stops periodic re-runs while everything still reports Healthy.
- **No surface.** The effective interval and where it came from appear only in logs.
- Carried from #877: the floor isn't operator-configurable, and failure retries are flat at the interval.

### Verification

`make lint` 0 issues, `make test` 15 suites. Reviewed for correctness, adversarially, and against #870's acceptance criteria — both Gherkin scenarios are proven by tests at both the Promise and resource-request layers.

### DIFF
--- a/api/v1alpha1/promiserevision_types.go
+++ b/api/v1alpha1/promiserevision_types.go
@@ -151,6 +151,9 @@ func (pr *PromiseRevision) ClearLatestRevisionLabel() {
 const MinReconciliationInterval = time.Minute
 
 // ReconciliationIntervalAnnotation overrides a revision's reconciliation interval.
+//
+// The revision for the Promise's current version mirrors this annotation from the Promise on
+// every reconcile. Revisions for older versions keep their annotation unchanged.
 const ReconciliationIntervalAnnotation = KratixPrefix + "reconciliation-interval"
 
 type ReconciliationIntervalSource string
--- a/api/v1alpha1/promiserevision_types.go
+++ b/api/v1alpha1/promiserevision_types.go
@@ -153,7 +153,8 @@ const MinReconciliationInterval = time.Minute
 // ReconciliationIntervalAnnotation overrides a revision's reconciliation interval.
 //
 // The revision for the Promise's current version mirrors this annotation from the Promise on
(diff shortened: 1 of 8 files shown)


aclevername (approved):
lgtm- but maybe another hint we should just remove promisereleases. took me too long to get my head round why we have this.

PR #887, opened 2026-08-24
fix: allow deleting a WorkPlacement that never wrote to the state store

### LINKED ISSUE #886: bug: WorkPlacement stuck `Pending` forever when a resource request is deleted before its first write
On a destination with `filepath.mode: none`, Kratix writes a small index file listing which files it put in the state store, so it knows what to remove later. Cleanup needs that index, but the index is only created by the first write, so deleting a resource request before that first write happens leaves the WorkPlacement undeletable: the request and the Work disappear, while the WorkPlacement keeps its deletionTimestamp, both finalizers and a `FailedReadKratixStateFile` event every 15 seconds, forever. You have to remove the finalizers by hand.

Cleanup reads the `.kratix` state file to know what to delete, and treats a missing file as an error ([workplacement_controller.go#L619](https://github.com/syntasso/kratix/blob/ae170e22e974e45bb7cfda00986057e01b3ccfde/internal/controller/workplacement_controller.go#L619)):

```go
stateFile, err := w.readKratixStateFile(repo, false)
```

The index may never turn up, because once something is marked for deletion Kratix stops writing to the state store, as every reconcile enters the deletion path before reaching the write path ([#L199](https://github.com/syntasso/kratix/blob/ae170e22e974e45bb7cfda00986057e01b3ccfde/internal/controller/workplacement_controller.go#L199)). Cleanup will then be stuck waiting for a file nothing can produce any more. It retries with a `nil` error, so there is no metric, backoff or warning log either.

The fix is to pass `true` and tolerate the missing file: no state file during deletion means that nothing was written, so there is nothing to clean up. Git commits the state file and the workloads together, and S3 writes the state file first ([#L560](https://github.com/syntasso/kratix/blob/ae170e22e974e45bb7cfda00986057e01b3ccfde/internal/controller/workplacement_controller.go#L560)), so it can never be missing while workloads exist.

To reproduce deterministically: create a request against such a destination, scale the manager to zero while the WorkPlacement is still unwritten, delete the request, scale back up.

This can also be triggered by a busy state store: writes to one state store happen one at a time, since every WorkPlacement backed by it takes the same lock for its whole reconcile, fetch included ([#L192](https://github.com/syntasso/kratix/blob/ae170e22e974e45bb7cfda00986057e01b3ccfde/internal/controller/workplacement_controller.go#L192)). A WorkPlacement queued behind other writes stays unwritten until its turn comes. On an idle platform that wait is a few seconds, but on a busy one it is long enough that a delete can be issued before the initial write, triggering the bug.

jeduardo:
On a `filepathMode: none` destination, cleanup reads the `.kratix` state file to know what to delete, and treats a missing file as an error. Only the first write creates that file, so a WorkPlacement deleted before its first write can never be removed. The resource request and the Work disappear normally, leaving nothing that points at the orphan.

A missing state file means the first write never happened, so there is nothing to clean up. We will accept deletions without the state file to avoid deadlocks, as the write to create the state file won't be hit after the resource is marked as deleted.

Closes #886

### DIFF
--- a/internal/controller/workplacement_controller.go
+++ b/internal/controller/workplacement_controller.go
@@ -616,7 +616,7 @@ func (w *workPlacementReconcileContext) deleteWorkPlacement(repo *Repository) (c
 
 		case v1alpha1.FilepathModeNone:
 			logging.Trace(w.logger, "handling file path mode none")
-			stateFile, err := w.readKratixStateFile(repo, false)
+			stateFile, err := w.readKratixStateFile(repo, true)
 			if err != nil {
 				logging.Debug(w.logger, "failed to read .kratix state file", "error", err)
 				return defaultRequeue, nil
--- a/internal/controller/workplacement_controller_test.go
+++ b/internal/controller/workplacement_controller_test.go
@@ -33,6 +33,7 @@ import (
 	"github.com/syntasso/kratix/api/v1alpha1"
 	"github.com/syntasso/kratix/lib/compression"
 	"github.com/syntasso/kratix/lib/hash"
+	"github.com/syntasso/kratix/lib/writers"
 	"github.com/syntasso/kratix/lib/writers/writersfakes"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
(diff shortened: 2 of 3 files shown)


jeduardo:
@codex review

jeduardo:
@codex review

senglezou (approved):

PR #893, opened 2026-08-26
fix: reap orphaned git helper processes with tini as PID 1

### LINKED ISSUE #892: bug: Every Git Fetch/Push leaves an orphaned PID, eventually causing PID exhaustion
## What happens

The `kratix-platform-controller-manager` pod slowly fills up with dead `git` processes. After a few days the pod reaches its process limit. Once that happens, it cannot start any new process at all.

You then see errors like this:

```
error: cannot fork() for remote-https: Resource temporarily unavailable
```

Writing to a Git state store stops working. `kubectl exec` into the pod also fails, because that needs a new process too.

The only way to recover today is to delete the pod:

```
kubectl delete pod -n kratix-platform-system <pod>
```

The new pod starts clean, and then the problem starts again.

This affects any Kratix installation that uses a Git state store. The more resource requests you have, the faster it fills up, because the leak grows with the number of reconciles.

## How to reproduce

Start a debug container alongside kratix, and observe the pid count go up everytime a Git workplacement is reconciled.
0. Setup env with a workplacement written to git
```
./scripts/quick-start.sh --git --recreate --single-cluster
k label destinations.platform.kratix.io platform-cluster environment=dev
kaf ~/workspace/kratix-marketplace/redis/promise.yaml
kaf ~/workspace/kratix-marketplace/redis/resource-request.yaml
k get workplacements -w # wait for workplacement to be created
```
1. Start a debug container (requires special security context since we run with no root allowed by default)
```
cat > /tmp/debug.json <<'EOF'
{
  "securityContext": {
    "runAsUser": 1000,
    "runAsGroup": 1000,
    "runAsNonRoot": true,
    "allowPrivilegeEscalation": false,
    "capabilities": { "drop": ["ALL"] }
  }
}
EOF

kubectl debug -it <POD NAME> \
  -n kratix-platform-system \
  --image=busybox --target=manager \
  --profile=general --custom=/tmp/debug.json -- sh
```
2. check the PID count:
```
ps -T
```
3. Modify the `.spec.size` of redis request from small <-> large to trigger a change in the workplacement
4. Wait few seconds for workplacement to get updated, then check the PID count again
```
ps -T
```
5. Observe its going up, and its all orphaned Git processes.

<details>
<summary> **Script to reproduce** </summary>

```bash
!/usr/bin/env bash
# Reproduce: does each WorkPlacement reconcile leak zombie git processes?
# Toggles redis spec.size small<->large and counts zombies in the manager's PID namespace.
set -euo pipefail

NS=${NS:-kratix-platform-system}
RR_NS=${RR_NS:-default}
RR=${RR:-example}
ROUNDS=${ROUNDS:-3}
WAIT=${WAIT:-10}

POD=$(kubectl -n "$NS" get pods -o name | grep controller-manager | head -1 | cut -d/ -f2)
echo "pod: $POD"

# The kratix image is alpine/git, so the manager container has busybox sh + /proc.
CTR=manager
if ! kubectl -n "$NS" exec "$POD" -c manager -- sh -c 'echo ok' >/dev/null 2>&1; then
  echo "no shell in manager; using ephemeral debug container" >&2
  CTR=zdebug
  kubectl -n "$NS" get pod "$POD" -o jsonpath='{.spec.ephemeralContainers[*].name}' | grep -qw "$CTR" \
    || kubectl -n "$NS" debug "$POD" --image=busybox --target=manager -c "$CTR" \
         --profile=general --attach=false -- sleep 9000 >/dev/null
  n=0
  until kubectl -n "$NS" exec "$POD" -c "$CTR" -- true 2>/dev/null; do
    n=$((n+1)); [ "$n" -gt 30 ] && { echo "debug container never became ready" >&2; exit 1; }
    sleep 2
  done
fi
echo "counting via: $CTR"

count() {
  kubectl -n "$NS" exec "$POD" -c "$CTR" -- sh -c \
    'z=$(cat /proc/*/status 2>/dev/null | grep -c "^State:.*Z"); \
     p=$(ls /proc | grep -c "^[0-9]*$"); \
     echo "zombies=$z pids=$p"'
}

printf '\nbaseline              %s\n' "$(count)"

i=0
while [ "$i" -lt "$ROUNDS" ]; do
  i=$((i+1))
  cur=$(kubectl -n "$RR_NS" get redis "$RR" -o jsonpath='{.spec.size}')
  [ "$cur" = small ] && new=large || new=small
  kubectl -n "$RR_NS" patch redis "$RR" --type=merge -p "{\"spec\":{\"size\":\"$new\"}}" >/dev/null
  sleep "$WAIT"
  printf 'round %d (%-5s->%-5s) %s\n' "$i" "$cur" "$new" "$(count)"
done

printf '\ndefunct processes:\n'
kubectl -n "$NS" exec "$POD" -c "$CTR" -- ps -T 2>/dev/null | grep '\[' || echo "  (none)"
```

</details>

Every reconcile that writes to a Git state store adds 2 more dead processes. They never go away.

# Engineering note

A "zombie" is a process that has finished, but nobody has collected its exit code yet. It keeps its process ID (PID) until someone does. On Linux, this job belongs to the parent process. If the parent has already gone, the job passes to PID 1.

What happens here:

1. Kratix runs `git fetch` and `git push`. These are direct children of `/manager`. Go waits for them correctly, so they are cleaned up.
2. Each of those git commands starts its own helper process, `git remote-https`, to talk to the remote over HTTPS.
3. **_Something is causing the Git helper to get left behind_**. The helper becomes an orphan, so Linux moves it under PID 1.
4. In our container, PID 1 is `/manager` itself. Go only cleans up processes that it started directly. It does not clean up orphans. So the helper stays as a zombie forever.
5. Each zombie keeps one PID slot. The container has a limit (cgroup `pids.max`). When that limit is full, `fork()` fails.

**Why is the helper left behind?** Git expects to be able to leave orphan processes around, and for `init` (PID 1) to clean them up. This is normal Git behaviour rather than a Git bug, and other projects hit the same thing ([Flux](https://github.com/weaveworks/flux/issues/484), [Gitea](https://github.com/go-gitea/gitea/issues/19077),
  [git-sync](https://github.com/kubernetes/git-sync/issues/55)).

Before #851 it was 1 per reconcile, because only `push` used the network. #851 added a `git fetch` to every reconcile, which doubled the rate.

# Options

1. **Add `tini` (or `dumb-init`) as the container entrypoint.** `/manager` then becomes PID 2, and `tini` becomes PID 1 and cleans up orphans. It is a small change in `Dockerfile`. The same change is needed in the SKE image. Recommended.
2. **Set `shareProcessNamespace: true` on the pod.** The Kubernetes `pause` container then becomes PID 1 and cleans up orphans. No image change is needed, but every container in the pod can then see and signal the other containers' processes.
3. **Clean up orphans inside the Go process** (`SIGCHLD` + `wait4`). Not recommended. It can take the exit code that `os/exec` is still waiting for, and break real git commands.

SaphMB:
Fixes #892

## What & why

Every `git fetch`/`git push` spawns a `git remote-https` helper process. When it outlives its immediate parent, it reparents to PID 1 — which in the controller-manager container was `/manager`, a Go binary that never reaps unknown children. The helper zombifies permanently. Enough Git-state-store reconciles eventually exhaust the pod's cgroup `pids.max`, and `fork()` starts failing everywhere, including `kubectl exec`. Today the only recovery is deleting the pod.

Fix: add `tini` as the container's PID 1, per the issue's own recommended option.

## What's in this PR

1. **`Dockerfile`** — installs `tini`, changes `ENTRYPOINT` to `["/sbin/tini", "--", "/manager"]`.
2. **`config/manager/manager.yaml`** and **`charts/kratix/templates/distribution.yaml`** — both hard-coded `command: [/manager]`, which in Kubernetes overrides the image's `ENTRYPOINT` entirely. An adversarial design read (dispatched before any code was written) caught that a Dockerfile-only fix would have shipped as a no-op for exactly this reason. Both now run `command: ["/sbin/tini", "--", "/manager"]`; `args` is untouched.
3. **`scripts/verify-zombie-reaping.sh`** (new, manual tool, not wired into CI) — proves the mechanism directly against the built image without a cluster: launches it with `--entrypoint` overridden to simulate the old vs. fixed manifests' `command:`, has it orphan a `sleep 3`, and checks whether it's reaped. I ran this myself: `plain` mode (old manifests) → 1 zombie, red. `tini` mode (fixed manifests) → 0 zombies, green.

## Ledger

| # | Criterion | State |
|---|---|---|
| AC1 | Container's PID 1 reaps orphaned `git remote-https` helpers so they don't accumulate as permanent zombies | **covered** — verified empirically (red/green above) |
| AC2 | Fix ships in the OSS `kratix-platform-controller-manager` image (Dockerfile + both deployment manifests in this repo) | **covered** |
| AC3 | Manager still starts non-root (65532:65532) and still receives/forwards SIGTERM promptly for graceful shutdown | **covered** — verified: `tini` under the manifests' actual `securityContext` (cap-drop ALL, no-new-privileges, default seccomp) still reaps; a direct child under tini stops in 0s on `docker stop` vs. 5s+SIGKILL for the same process as bare PID 1 |
| AC4 | A regression check proves the orphan-reaping mechanism, not just that the image builds | **covered** — `scripts/verify-zombie-reaping.sh`, run manually (not CI-gated — see review note below) |
| AC5 | Same fix reaches every deployment path | **deferred** — two more repos found with the identical `command: [/manager]` override bug during the design read: the actual production Helm chart (`syntasso/helm-charts` `kratix/files/deployment.yaml` — the copy inside *this* repo, `charts/kratix/templates/distribution.yaml`, is a vestigial pre-`Chart.yaml` file CI overwrites before the e2e test runs, fixed here anyway since it's still used by local runs) → syntasso/helm-charts#234. The SKE platform image (`syntasso/enterprise-kratix`, different base image, `ENTRYPOINT []`, no `apk`) → syntasso/enterprise-kratix#1411. Both confirmed via inspection, not blind-guessed. |

## Process notes (delivery record)

- **No ADR governs this** — checked all 14 ADRs in `syntasso/adrs`; none address container entrypoints or the controller-manager image. Delivered as an ungrounded infra reliability fix.
- **Persona:** Penny, Platform Engineer — this bug is a direct regression of her "Efficiency (Fleet Management)" gain (manual pod-delete recovery, repeated forever, worse at scale).
- **Release level: unreleased** — no epic file, no ADR-derived level → branch + PR, no tag.
- **No pair assigned** — `~/.git-coauthors` doesn't exist on the delivering machine; proceeding solo was a mechanics decision, not a scope one.
- **`scripts/verify-zombie-reaping.sh` is not wired into CI in this PR.** CI's gate (`syntasso/ci` `test-kratix.yaml`) lives in a separate repo; cross-repo CI wiring is a bigger step than this story. A discrimination review confirmed every branch in the script is currently only exercised by a human running it — that's accepted, not a defect, given the above.
- **Gate run locally**, same SHA as pushed (`62177fba`, no rebase since): `make test` (15 suites, green — unchanged from the untouched-`main` baseline, no Go code touched), `golangci-lint run --config=.golangci-required.yml` at CI's pinned v2.11 (identical single pre-existing finding in `util/git/git.go:211`, unrelated, present in the baseline too), `docker build .`, `kustomize build config/default` (renders the fixed `command:` correctly). The kind-cluster-based `system-test`/`core-test`/`integration-test-*` CI jobs were not run locally; CI will run them on this PR.
- **Two independent review passes** (adversarial + the required discrimination pass) ran against the pinned commit SHA before this PR was opened. Adversarial pass empirically re-ran the verification script against a rebuilt image, including under the manifests' actual `securityContext` restrictions — no critical findings. Discrimination pass confirmed the script's branches are manual-only, as expected.


aclevername (approved):
lgtm!

PR #899, opened 2026-09-01
fix: clear stale Reconciled reason on promises with no configure workflow

kirederik:
## What's broken

Install a compound promise before the promises it requires, and it keeps saying its requirements are missing forever, long after they are installed and everything else about it looks healthy.

## How to see it

**1. Install a promise that requires another one, without installing what it requires.**

```
kubectl apply -f test/system/assets/compound-promise/promise.yaml
kubectl get promise compound-promise
```

It reports `Unavailable`, and its `Reconciled` condition says `RequirementsNotFulfilled`. Both correct — the promise it needs isn't there yet.

**2. Now install the promise it needs.**

```
kubectl apply -f test/system/assets/compound-promise/sub-promise.yaml
```

**3. Wait for it all to settle, then look again.**

```
kubectl get promise compound-promise -o yaml
```

- `sub-promise` — Available ✅
- `compound-promise` — Available ✅
- `RequirementsFulfilled` — True ✅
- `requiredPromises[].state` — Requirement installed ✅
- `Reconciled` — **still `RequirementsNotFulfilled`** ❌

## Why it happens

Kratix decides whether a promise is reconciled by checking whether that promise's own configure workflow finished.

In cases where a compound promise doesn't have one, Kratix never revisits `Reconciled`, and whatever it wrote there first is what stays. So a Promise that was in RequirementsNotFulfilled state would never change.

## Second commit

A tidy-up of the code that decides all this: it was a five-branch chain that worked out the answer and separately hand-checked whether the answer had changed, in the same breath.

### DIFF
--- a/api/v1alpha1/promise_types.go
+++ b/api/v1alpha1/promise_types.go
@@ -37,17 +37,18 @@ import (
 )
 
 const (
-	PromiseStatusAvailable               = "Available"
-	PromiseStatusUnavailable             = "Unavailable"
-	PromisePlural                        = "promises"
-	KratixResourceHashLabel              = "kratix.io/hash"
-	KratixPipelineHashLabel              = "kratix.io/pipeline-hash"
-	PromiseAvailableConditionType        = "Available"
-	PromiseAvailableConditionTrueReason  = "PromiseAvailable"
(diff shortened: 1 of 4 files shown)


SaphMB (approved):
Reproduced bug and validated fix - happy to approve

PR #908, opened 2026-09-03
fix: only aggregate health records from the same promise

### LINKED ISSUE #907: bug: HealthRecords leak across promises when resources share a name and namespace
HealthRecords are matched to resources by resourceRef.name and resourceRef.namespace only, while promiseRef.name is ignored:

https://github.com/syntasso/kratix/blob/main/internal/controller/healthrecord_controller.go#L147-L153

If two promises each have a resource with the same name in the same namespace, both resources show both promises' health records, and each one's aggregate state is affected by the other's. The delete path in deleteHealthRecord() rebuilds the status with no resource filter at all.

Expected: the filter also matches on promiseRef.name, in both places, avoiding wrong aggregation of health statuses.

jeduardo:
HealthRecords were matched to resources by resourceRef name and namespace only, so resources from different promises sharing a name and namespace absorbed each other's health records. Match on promiseRef.name as well, in both the aggregation and delete paths.

Fixes #907

### DIFF
--- a/internal/controller/healthrecord_controller.go
+++ b/internal/controller/healthrecord_controller.go
@@ -146,8 +146,7 @@ func (r *HealthRecordReconciler) updateResourceStatus(
 
 	var resourceHealthRecords []platformv1alpha1.HealthRecord
 	for _, record := range healthRecords.Items {
-		if record.Data.ResourceRef.Name == healthRecord.Data.ResourceRef.Name &&
-			record.Data.ResourceRef.Namespace == healthRecord.Data.ResourceRef.Namespace {
+		if referToSameResource(&record, healthRecord) {
 			resourceHealthRecords = append(resourceHealthRecords, record)
 		}
 	}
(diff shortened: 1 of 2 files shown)


internal/controller/healthrecord_controller.go
```diff
@@ -222,6 +221,15 @@ func (r *HealthRecordReconciler) getInitialHealthStatusState(resReq *unstructure
 	return initialHealthStatusState
 }
 
+// referSameResource reports whether two health records target the same resource request.
+// The promise name is part of the identity: resources from different promises can share a
+// name and namespace without sharing health records.
+func referSameResource(a, b *platformv1alpha1.HealthRecord) bool {
```
jeduardo:
```suggestion
func referToSameResource(a, b *platformv1alpha1.HealthRecord) bool {
```

internal/controller/healthrecord_controller.go
```diff
@@ -291,7 +296,7 @@ func (r *HealthRecordReconciler) deleteHealthRecord(
 	}
 
 	for _, record := range healthRecords.Items {
-		if record.GetName() != healthRecord.GetName() {
+		if record.GetName() != healthRecord.GetName() && referToSameResource(&record, healthRecord) {
```
kirederik, line 299: nit: i guess this is to avoid updating unrelated records; can we filter when we list instead?

kirederik (approved):

internal/controller/healthrecord_controller.go
jeduardo: Raised a separate follow up to make this better https://github.com/syntasso/kratix/issues/909

PR #916, opened 2026-09-09
test: de-serialise the system suite via destination isolation

jeduardo:
16 of the 25 specs in `test/system` are `Serial`, forming a ~34 minute tail that is most of the suite's wall clock. They are Serial to stop specs leaking into each other: a promise without `destinationSelectors` schedules dependency works to every destination and resource works to one at random, so any spec's destination could receive another spec's files.

This swaps that protection for isolation. Every test destination gets `strictMatchLabels` and a unique label, so selector-less works can never land on it, and the specs Serial was guarding rejoin the parallel pool. What stays Serial genuinely conflicts: the specs that swap platform-wide config or patch shared state store credentials.

## What changed

**Destination isolation** (`test/system/assets/destination/*.yaml`) — `strictMatchLabels` plus a unique label on each test destination; the two status destinations also lose their shared `environment: dev` label. New `destination-git-recovery-store.yaml` gives git-recovery its own state store.

**Specs out of `Serial`** — filepath modes, cleanup-all (`destination_test.go`), rbac, dry-run flag, resource-binding-latest (`resource_binding_default_version_test.go`), git-recovery (`git_recovery_test.go`, now pushing to its own Gitea repo). `compound_promise_test.go` becomes `Ordered` rather than parallel: its two specs share promise names.

**Long specs split in half** (`workflow_control_test.go`, `promise_revision_test.go`, plus the new `*-del.yaml` / `*-fail.yaml` fixtures) — each half gets its own promise fixtures so they are independent. The suite's straggler drops 315s → 223s, which is the floor for every future shard.

**Transient error retries** (`test/kubeutils/cluster.go`, `scripts/quick-start.sh`) — `kubectl` retries TLS handshake timeouts, connection resets and etcd timeouts before asserting, and quick-start retries image loads while containerd starts. Both are insurance for loaded runners rather than fixes for anything failing today.

**Two latent test bugs** the parallel pool exposed — the `filepathmode none` spec depended on a Serial spec having applied its state store, and asserted on an unscoped workplacement listing; `promise_revision` deleted resource bindings without scoping to its own promise.

## Results

Measured on this branch in CI: `system-test` drops from 42.8/44.8 to 32.1/31.0 minutes (bucket/git), 28/28 green — [run 34115634519](https://github.com/syntasso/ci/actions/runs/34115634519). Coverage compared before and after across instrumented runs: net one statement of 8,261.

Part of syntasso/enterprise-kratix#1382, which lists `system-test` not running in parallel as its second consideration. First of a short series — the next PRs cut redundant controller restarts, shard the job across runners, and pick the changes up in the SKE pipeline.


test/system/workflow_control_test.go
```diff
@@ -319,113 +355,137 @@ var _ = Describe("Workflow Control", func() {
...
+		})
+	})
 
+	When("resource and delete pipelines are suspended by the workflow control file", func() {
+		BeforeEach(func() {
+			// clear all suspend gates in case a previous run was interrupted
+			platform.Kubectl("delete", "-f", suspendDelConfigMap, "--ignore-not-found")
+			platform.Kubectl("delete", "cm", suspendDelPromiseDeleteGate, "-n", "kratix-platform-system", "--ignore-not-found")
+
+			// This promise copy has no configure-time suspensions, so it installs clean.
```
ChunyiLyu:
```suggestion
```

test/system/assets/workflow-control/promise-retry-del.yaml
```diff
@@ -0,0 +1,164 @@
...
+            name: pipe-2
+          spec:
+            containers:
+              - name: complete
+                image: ghcr.io/syntasso/kratix-pipeline-utility:v0.0.1
+                command: ["/bin/sh", "-c"]
+                args:
+                  - |
+                    set -eu
+                    echo "pipe-2 complete"
```
ChunyiLyu: Nothing in this spec asserts on the promise configure pipelines. Think we can remove it.

```suggestion
```

test/system/assets/workflow-control/promise-retry-del.yaml
```diff
@@ -0,0 +1,164 @@
...
+                  - |
+                    set -eu
+                    if kubectl get configmap workflow-retry-del-resource-gate -n default >/dev/null 2>&1; then
+                      echo "gate exists; configure complete"
+                      exit 0
+                    fi
+                    cat <<'EOF' > /kratix/metadata/workflow-control.yaml
+                    retryAfter: 10s
+                    message: retrying resource pipeline
+                    EOF
```
ChunyiLyu: Think we don't need to have wait and trigger behavior in configure workflow for a test asset on delete test. Similarly, we can also drop the configmap permission.

```suggestion
            containers:
              - name: complete
                image: ghcr.io/syntasso/kratix-pipeline-utility:v0.0.1
                command: ["/bin/sh", "-c"]
                args: ["echo 'configure complete'"]
```

test/system/assets/workflow-control/promise-suspend-del.yaml
```diff
@@ -0,0 +1,164 @@
...
+            name: pipe-2
+          spec:
+            containers:
+              - name: complete
+                image: ghcr.io/syntasso/kratix-pipeline-utility:v0.0.1
+                command: ["/bin/sh", "-c"]
+                args:
+                  - |
+                    set -eu
+                    echo "pipe-2 complete from v2"
```
ChunyiLyu: similar as above

```suggestion
```

ChunyiLyu (commented):
A couple of suggestions on the workflow control test on delete pipelines. Looks good!

test/system/workflow_control_test.go
```diff
@@ -121,56 +128,89 @@ var _ = Describe("Workflow Control", func() {
...
+			Eventually(func(g Gomega) {
+				g.Expect(platform.Kubectl("get", "workflowretrydels", "retry-test-del", phaseJSONPath("resource-pipe-retry"))).To(Equal("Succeeded"))
+			}).Should(Succeed())
+		})
+
+		AfterEach(func() {
+			platform.KubectlAllowFail("create", "cm", retryDelResourceDeleteGate, "-n", "default")
+			platform.KubectlAllowFail("create", "cm", retryDelPromiseDeleteGate, "-n", "kratix-platform-system")
+			platform.EventuallyKubectlDelete("promise", retryDelPromiseName, "--ignore-not-found")
+			platform.Kubectl("delete", "cm", retryDelDependentCM, "-n", "kratix-platform-system", "--ignore-not-found")
```
ChunyiLyu:
```suggestion
```
I think we can remove this line now given I deleted the configure workflow logic in file `promise-retry-del.yaml`

test/system/workflow_control_test.go
```diff
@@ -121,56 +128,89 @@ var _ = Describe("Workflow Control", func() {
...
 				}).Should(Succeed())
 			})
+		})
+	})
 
+	When("the delete pipelines have 'retryAfter' set", func() {
+		BeforeEach(func() {
+			// open the configure gates upfront so setup completes without retries,
+			// and clear any leftover delete gates
+			platform.KubectlAllowFail("create", "cm", retryDelDependentCM, "-n", "kratix-platform-system")
```
ChunyiLyu:
```suggestion
```
I think we can remove this line now given I deleted the configure workflow logic in file `promise-retry-del.yaml`

PR #918, opened 2026-09-11
fix: stop two deleting HealthRecords from livelocking each other

jeduardo:
A HealthRecord drops its finalizer once it is absent from its resource's `status.healthStatus.healthRecords`. `deleteHealthRecord` rebuilt that list from every other record for the same resource, including ones also being deleted, so each pass put the other back.

Two records for one resource deleted together therefore never saw themselves absent, and both sat `Terminating` indefinitely until a finalizer was removed by hand. Resources with a single record deleted cleanly, which is the tell.

Records being deleted are now excluded from the rebuilt list.

The test alternates single reconcile passes between the two records rather than using `reconcileUntilCompletion`: completing one tidies up before the other starts, which hides the bug. Without the fix, only this test fails.


internal/controller/healthrecord_controller.go
```diff
@@ -296,6 +296,18 @@ func (r *HealthRecordReconciler) deleteHealthRecord(
...
 	for _, record := range healthRecords.Items {
+		// Records already being deleted are left out, not just this one.
+		//
+		// The finalizer is dropped on the pass that finds this record absent
+		// from the resource's list, so whatever is written here has to be a list
+		// this record can disappear from. Including another deleting record puts
+		// *it* back, and when two records for one resource are deleted together
+		// each pass re-adds the other: both always find themselves listed,
+		// neither ever reaches the branch above, and the two livelock until
+		// something removes a finalizer by hand.
```
jeduardo: Make this comment simpler, two lines maximum

internal/controller/healthrecord_controller_test.go
```diff
@@ -376,6 +377,71 @@ var _ = Describe("HealthRecordController", func() {
...
+
+		// Each record drops its finalizer on the pass that finds it absent from
+		// the resource's list. Rebuilding that list from the other deleting
+		// record puts it straight back, so with the two interleaving — which is
+		// what the controller does, one pass each in turn — neither is ever
+		// absent and both spin forever.
+		//
+		// Deliberately alternating single passes rather than driving each record
+		// to completion: run one to completion first and it tidies up before the
+		// other starts, which is exactly why this went unnoticed.
```
jeduardo: Make this simpler, two lines maximum

internal/controller/healthrecord_controller.go
jeduardo: Done — cut to two lines.

internal/controller/healthrecord_controller_test.go
jeduardo: Done — cut to two lines.

internal/controller/healthrecord_controller_test.go
```diff
@@ -434,6 +435,64 @@ var _ = Describe("HealthRecordController", func() {
...
+			for pass := 0; pass < 20 && (!gone(firstName) || !gone(secondName)); pass++ {
+				for _, name := range []types.NamespacedName{firstName, secondName} {
+					if gone(name) {
+						continue
+					}
+					_, err := reconciler.Reconcile(ctx, ctrl.Request{NamespacedName: name})
+					Expect(err).NotTo(HaveOccurred())
+				}
+			}
+
```
aclevername: For loops in tests are a bit of a smell IMO. Could this be an Eventually and/or Consistently?

aclevername (approved):
1 nit - bug fix is good to merge

it does feel like a simpler model might be the Resource Request buildings its own status's healthrecord list, rather than each healthrecorud pushing/deleting themselves from the list, but thats a bigger refactor thats not worth doing now.

internal/controller/healthrecord_controller_test.go
jeduardo: Good catch, switched to an Eventually!

PR #921, opened 2026-09-14
test: cut redundant controller restarts and padded waits

jeduardo:
Every config-mutating spec restored the default Kratix config and restarted the controller in `AfterEach` — even though the next spec applies its own config in `BeforeEach` anyway. That is 8 restarts per run where 4 will do.

Follows #916. Now that the suite runs in parallel, what is left in the Serial tail is config-mutating work, so this makes that tail as short as the restarts allow.

## What to look at

| File | Change |
| --- | --- |
| `system_suite_test.go` | Restore the config once, in `SynchronizedAfterSuite`. Also gives the manager 2 CPU / 1Gi; it ships with 100m / 256Mi and throttles under parallel load. |
| `reconcile_after_failure_test.go` | **Largest diff; each `When` moves unchanged.** Describes regrouped by config rather than by subject, so `BeforeAll` applies each config once. |
| `resource_binding_default_version_test.go` | Drop the now-redundant restore and restart. |
| `reconciliation_test.go` | Promise-pause `Consistently` 70s → 30s; reconciliation there is event-driven. |
| `test/kubeutils/cluster.go` | `EventuallyKubectlDelete` outer poll 1ms → configured interval. |
| `Makefile` | `GINKGO_PROCS` to cap procs locally; `-p` overloads a laptop's kind apiserver. |

Net effect: 8 restarts per run become 4 — two `Ordered` groups, the suite restore, and BeforeSuite.

## Results

`system-test` drops from 32.1/31.0 to **28.7/30.1 minutes** (bucket/git) on top of #916 — best of two green runs, [one](https://github.com/syntasso/ci/actions/runs/34115637521) on the branch and [one](https://github.com/syntasso/kratix/actions/runs/34854094666) on this PR. Against the pre-#916 baseline of 42.8/44.8 minutes, that is about a third off the job.

Part of syntasso/enterprise-kratix#1382, second of the series.


test/system/reconcile_after_failure_test.go
```diff
@@ -104,206 +126,199 @@ var _ = Describe("Reconcile after failure", Serial, func() {
...
+})
+
+var _ = Describe("Reconcile after failure disabled", Label("config-mutating"), Serial, Ordered, func() {
+	BeforeAll(func() {
+		rafSetTimeouts()
+		platform.Kubectl("apply", "-f", filepath.Join(rafAssetsPath, "kratix-config-no-retry.yaml"))
+		restartController()
+	})
+
+	BeforeEach(rafSetTimeouts)
```
kirederik: called twice? first time on the BeforeAll above

test/system/resource_binding_default_version_test.go
```diff
@@ -66,12 +66,11 @@ var _ = Describe("ResourceBinding Default Version pinned", Label("config-mutatin
 		}).Should(ContainSubstring("Available"))
 	})
 
+	// The default config is not restored here: the suite restores it once at
+	// the end, after all config-mutating Serial specs have run.
```
kirederik: this comment only makes sense when looking at the diff. In two weeks we'll be like "wat"

test/system/reconciliation_test.go
```diff
@@ -62,9 +62,12 @@ var _ = Describe("Reconciliation", func() {
 			}).Should(ContainSubstring("Paused"))
 
 			By("not running any workflow while paused")
+			// Reconciliation is event-driven here (the global re-reconcile
+			// interval is hours), so a short window is enough to catch a
+			// wrongly-triggered workflow.
```
kirederik: this comment only makes sense when looking at the diff. In two weeks we'll be like "wat"

kirederik (approved):

kirederik (approved):

test/system/reconcile_after_failure_test.go
jeduardo: Thanks, removed!

test/system/reconciliation_test.go
jeduardo: Thanks, reworded it!

test/system/resource_binding_default_version_test.go
jeduardo: Also dropped it

PR #933, opened 2026-09-15
chore: resolve workflow progression from pipeline status, not jobs

aclevername:
Use keyed pipeline status and a single hash to advance configure and delete workflows, recreate missing Jobs, and retain progress through Job cleanup. Migrate legacy resource status and share the reconciliation path.

Promise workflows run once after upgrading. Controllers that embed the workflow engine select their own status key through WorkflowKey.


config/crd/bases/platform.kratix.io_promises.yaml
```diff
@@ -257,36 +257,44 @@ spec:
                       if installed via one
                     type: string
                   workflows:
+                    additionalProperties:
```
aclevername, line 260: wtf?

aclevername:
@codex review

aclevername:
@claude review

PR #936, opened 2026-09-16
feat: set a content type on objects written to a bucket

jeduardo:
Objects written to a bucket state store carry no content type, so they land as `application/octet-stream`. This sets the type from the file extension: `application/yaml` for `.yaml` and `.yml`, `application/json` for `.json`, and the store's default for anything else.

Refs https://github.com/syntasso/enterprise-kratix/issues/1529


jeduardo:
Abandoned: it doesn't solve the browser problem.

Issue #811, opened 2026-06-02
chore: decouple workflow progress from Job objects


kirederik:
### Context

The configure workflow engine uses **existing Kubernetes Job objects** to decide which pipeline to run next. When Jobs are removed, `determineWorkflowState` resets to pipeline 0 even when `status.kratix.workflows.pipelines` already records progress.

That makes aggressive Job cleanup unsafe. Parent status already tracks per-pipeline phases and the `ConfigureWorkflowCompleted` condition; the workflow library should use those as the source of truth, not retained Jobs.

### Engineering notes

- **Workflow reconciler** (`lib/workflow/reconciler.go`): `determineWorkflowState`, `executeReconcileAction`, and `handleCurrentPipelineJob` are Job-centric today.
- **Parent status**: `status.kratix.workflows.pipelines` and workflow conditions drive progression; deprecated workflow counters should not be updated.
- **Re-runs**: Controllers already re-trigger configure workflows on spec change or after `reconciliationInterval`; this must still work when prior Jobs are gone.

### Acceptance criteria

```gherkin
Feature: Status-driven configure workflow progression

Background:
  Given a Promise with multiple configure pipelines
  And a resource request for that Promise

Scenario: Advance when Jobs are gone but status shows progress
  Given pipeline 0 is Succeeded in status.kratix.workflows.pipelines
  And no Jobs remain for this workflow
  When the resource request is reconciled
  Then a Job is created for pipeline 1
  And pipeline 0 is not run again

Scenario: Re-run after full success when a new run is required
  Given all configure pipelines are Succeeded
  And the ConfigureWorkflowCompleted condition is True
  And no Jobs remain for this workflow
  And the resource request spec has changed or the reconciliation interval has elapsed
  When the resource request is reconciled
  Then a Job is created for pipeline 0
  And status reflects a new configure run in progress

Scenario: In-flight Job prevents duplicate creation
  Given a Job for the current pipeline is still Running
  When the resource request is reconciled
  Then no second Job is created for that pipeline
```

### Tasks

- [x] Refactor `determineWorkflowState` to derive pipeline index from status, not Job history
- [x] Refactor `executeReconcileAction` / `handleCurrentPipelineJob` to start the next (or re-run) pipeline when status requires it, even if `mostRecentJob` is nil
- [x] Add reconciler tests for: advance with no Jobs; re-run after success with no Jobs; no duplicate while Running
- [x] Run existing workflow and resource-request tests; fix Job-retention assumptions
[...]

PR #937, opened 2026-09-17
chore: Cleanup jobs that don't belong to any existing pipeline as part of cleanup

### LINKED ISSUE #810: bug: jobs not cleaned up when workflow does not complete
## Context

Job cleanup only executes when the final pipeline completes successfully: https://github.com/syntasso/kratix/blob/385cd6da687a2aa06e89df0aa96a8d3135a05423/lib/workflow/reconciler.go#L356-L372. 

This means jobs accumulate without bound when:

  1. The middle or the last pipeline has a long time out and does not complete
  2. A promise updates, or reconciliation interval triggers
  3. kratix suspends the running last pipeline and starts from the first pipeline
  4. no job clean up ever triggered

You ended up accumulating jobs and they exceed the `numberOfJobsToKeep` limit.

### Expected behaviour

Job count per pipeline per resource should respect `numberOfJobsToKeep` regardless of workflow interruptions or failures.

aclevername:
Closes #810.

See comment: https://github.com/syntasso/kratix/issues/810#issuecomment-5712882647
> This issue got partially (accidentally) addressed in https://github.com/syntasso/kratix/issues/811, we added a bunch of new cleanup calls, including ending every create pipeline call with a call to cleanup jobs. When running this issue throug with Claude, the only thing we spotted is that we wouldn't cleanup jobs belonging to pipelines that have been removed/renamed, so we've ammended that instead


lib/workflow/reconciler.go
```diff
@@ -441,25 +441,38 @@ func cleanup(opts Opts, namespace string) error {
...
+	if err != nil {
+		logging.Error(opts.logger, err, "failed to list jobs for workflow")
+		return err
+	}
+
+	for pipelineName, jobsForPipeline := range groupJobsByPipelineName(jobsForWorkflow) {
 		if err := pruneJobs(opts, jobsForPipeline); err != nil {
-			logging.Error(opts.logger, err, "failed to delete old jobs")
+			logging.Error(opts.logger, err, "failed to delete old jobs", "pipeline", pipelineName)
 			return err
```
kirederik: does this need to be terminal (i.e. return err)? can we try to delete the jobs we can delete and move on? I know this was not introduced as part of this pr, but...

lib/workflow/reconciler.go
```diff
@@ -441,25 +441,38 @@ func cleanup(opts Opts, namespace string) error {
...
 		}
 	}
 
 	return nil
 }
 
+func groupJobsByPipelineName(jobs []batchv1.Job) map[string][]batchv1.Job {
```
kirederik, line 470: it looks like we introduced this function _just_ to get the pipeline name in the logging? I'd reckon the name of the job that we failed to delete is more useful information, which would then remove the need for this function entirely?

kirederik (approved):
a couple of nits, but overall ok

lib/workflow/reconciler.go
aclevername: not quite, its to ensure we delete jobs who belong to pipelines that don't exist anymore< those pipeline names are only discoverable on the jobs, not in the promise

lib/workflow/reconciler.go
aclevername: its a fair comment and i agree, woll update

PR #938, opened 2026-09-17
feat: configure workflow job TTL

aclevername:
Done as part of https://github.com/syntasso/kratix/issues/810

Now we no longer require jobs to exist to track the history of pipeline runs, we can support users configuring TTL of them

default behavior is unchanged, jobs live forever until pruned


aclevername:
@codex review

api/v1alpha1/pipeline_types.go
```diff
@@ -138,6 +141,8 @@ type Permission struct {
 type JobOptions struct {
 	// Number of retries before marking the pipeline Job as failed
 	BackoffLimit *int32 `json:"backoffLimit,omitempty"`
+	// Number of seconds to retain the pipeline Job after it finishes
+	TTLSecondsAfterFinished *int32 `json:"ttlSecondsAfterFinished,omitempty"`
```
jeduardo, line 145: We are emitting warnings about this value and clamping it in three different locations. Is there a way to do this in a single place? Maybe a validating or mutating webhook?

jeduardo (approved):
It looks fine, but I wonder if there is a way to make it even simpler (comment in-line).

api/v1alpha1/pipeline_types.go
aclevername: Part of this is because it comes from 2 different sources, so i don't think we'd get to just 1 place

i have deleted 1 of the checks that was redundant

added webhook to validate the promise inlined configuration

configmap approach stays as is

PR #943, opened 2026-09-18
feat: allow TLS validation to be enabled on GitStateStore

jeduardo:
Adds `spec.insecure` to GitStateStore so TLS validation can be turned on for HTTPS git servers. It was hardcoded to `true` in the writer, so certificates were never validated.

Defaults to `true`, so nothing changes for existing stores.

SSH is unaffected.


api/v1alpha1/gitstatestore_types.go
```diff
@@ -34,6 +34,12 @@ type GitStateStoreSpec struct {
 	// URL of the git repository.
 	URL string `json:"url,omitempty"`
 
+	// Toggle to turn off or on TLS verification when connecting to the repository.
+	// Only applies to HTTPS connections.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default:=true
```
aclevername: an inverted word `insecure` having a default of `true` is bit smelly.

`secure: false` would be less smelly yaml wise- but breaks conventions in general of `insecure` being the word to relate to skipping tls


could we do `insecure: false # default` instead? i guess its too big a breaking change, but does feel weird that we are defaulting to insecure true, both from a yaml smell and general best practices point of view.

api/v1alpha1/gitstatestore_types.go
```diff
@@ -34,6 +34,12 @@ type GitStateStoreSpec struct {
...
+	Insecure bool `json:"insecure"`
```
aclevername: do we need to provide a way for them to provide custom CA? i can imagine if they are using a self hosted git provider this is probably mandatory. not everyone will have them pre-injected into the containers CA pool.

i think this is mandatory to meaningfully deliver this feature

config/crd/bases/platform.kratix.io_gitstatestores.yaml
```diff
@@ -81,6 +81,12 @@ spec:
                     description: Name of the Git commit author
                     type: string
                 type: object
+              insecure:
+                default: true
+                description: |-
+                  Toggle to turn off or on TLS verification when connecting to the repository.
+                  Only applies to HTTPS connections.
```
aclevername: so is the variable only used when `.spec.url` is prefixed with `https`? should we have the validating webhook warn/error if a user sets `insecure: false` with a url that isn't `https`? just to make sure they don't think its doing something else.

might be overkill

api/v1alpha1/gitstatestore_types.go
aclevername: Do we want this default to only apply in our code? If a user never set `insecure: true`, seeing it come back in the yaml could be confusing. Over ssh it does nothing anyway. It's only a useful stored default when the url is `https`.

test/system/destination_test.go
```diff
@@ -97,6 +97,18 @@ var _ = Describe("Destinations", Label("destination"), Serial, func() {
 			WaitReady("destination", destinationName)
 
 			if os.Getenv("LRE") != "true" {
+				// gitea serves a self-signed cert, so validating it must fail
+				By("failing when TLS validation is enabled", func() {
+					platform.Kubectl("patch", "gitstatestore", stateStoreName, "--type=merge", "-p", `{"spec":{"insecure":false}}`)
+
+					ExpectNotReady("gitstatestore", stateStoreName)
+					ExpectEvent("gitstatestore", stateStoreName, "unable to clone repository", "certificate")
```
aclevername: could we make this assertion more explicit? `certficate` is a bit of a vague assertion

config/crd/bases/platform.kratix.io_gitstatestores.yaml
jeduardo: I'll raise a warning for the situation of insecure: true + ssh, good catch

api/v1alpha1/gitstatestore_types.go
jeduardo: That's a good point, I don't think it makes sense for it to show up if the user doesn't set it. I'll check how to do it with webhooks.

jeduardo: Changed to use a webhook that issues a warning and having the code to default to insecure = false without surfacing the field in the yaml when the user doesn't specify it.

test/system/destination_test.go
jeduardo: Added a better message: TLS certificate validation failed

internal/controller/repository_cache.go
```diff
@@ -13,6 +14,27 @@ import (
...
 )
 
+// tlsValidationFailures are how git and the Go HTTP client report a server certificate
+// they cannot verify. The underlying cause varies (self-signed, unknown CA, expired), and
+// the wording differs between the openssl and gnutls git builds.
+var tlsValidationFailures = []string{
+	"SSL certificate problem",
+	"server certificate verification failed",
+	"x509: certificate signed by unknown authority",
+	"tls: failed to verify certificate",
```
aclevername: are these errors from underlying libraries? If we can't import them as an error obj, i dont think hard code copying them is best choice here. smells

internal/webhook/v1alpha1/gitstatestore_webhook.go
```diff
@@ -0,0 +1,59 @@
...
+// is only ever negotiated over https.
+func warnOnInsecureWithoutHTTPS(gitStateStore *v1alpha1.GitStateStore) admission.Warnings {
+	url := gitStateStore.Spec.URL
+	if gitStateStore.Spec.Insecure == nil || strings.HasPrefix(strings.ToLower(url), "https://") {
+		return nil
+	}
+
+	return admission.Warnings{
+		fmt.Sprintf("spec.insecure only applies to https urls; it has no effect on %s", url),
+	}
```
aclevername: i think generally in code base we are a fan of early `return nil` instead of complex if's, but here its just the 1 if, so having an inverted early return "don't validate if X or Y" is worse to read than "Validate if X and Y".


```suggestion
	if gitStateStore.Spec.Insecure != nil && !strings.HasPrefix(strings.ToLower(url), "https://") {
		return admission.Warnings{
			fmt.Sprintf("spec.insecure only applies to https urls; it has no effect on %s", url),
	    }
	}

	return nil
```

aclevername (approved):
left 2 nit's, and also noticed CI red. apart fromt hat LGTM

config/crd/bases/platform.kratix.io_gitstatestores.yaml
jeduardo: Resolving this now as the webhook is raising the warning

internal/controller/repository_cache.go
jeduardo: Removed this

internal/webhook/v1alpha1/gitstatestore_webhook.go
jeduardo: I'm fine with doing it this way, just took it

